PR #33, opened 2020-02-09
removed defer in String func in value.go

selam:
don't use defer on unsafe package.


selam:
I was talk too soon, after running for some time (1.5 hour later) i get same crash and i can't find why, so i decided to recover crashed goroutine and run again.  sorry for miss leading.

ghoshabhi:
Can you close this PR in that case @selam ? :)

PR #40, opened 2020-11-27
context: Add print to context globals.

tmc:
This is a the first step in supplying some bi-directional interoperability between the v8 isolates and Go.

### DIFF
--- a/v8go.cc
+++ b/v8go.cc
@@ -134,6 +134,24 @@ IsolateHStatistics IsolationGetHeapStatistics(IsolatePtr ptr) {
 
 /********** Context **********/
 
+// Print provides a basic interface to print to stdout.
+static void Print(const FunctionCallbackInfo<Value>& args) {
+  bool first = true;
+  auto iso = args.GetIsolate();
+  for (int i = 0; i < args.Length(); i++) {
+    HandleScope handle_scope(iso);
+    if (first) {
+      first = false;
+    } else {
+      printf(" ");
+    }
+    String::Utf8Value value(iso, args[i]);
+    printf("%s", *value);
+  }
+  printf("\n");
+  fflush(stdout);
+}
+
 ContextPtr NewContext(IsolatePtr ptr) {
     Isolate* iso = static_cast<Isolate*>(ptr);
     Locker locker(iso);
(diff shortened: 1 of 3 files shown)


v8go.cc
```diff
@@ -134,16 +134,38 @@ IsolateHStatistics IsolationGetHeapStatistics(IsolatePtr ptr) {
...
 ContextPtr NewContext(IsolatePtr ptr) {
     Isolate* iso = static_cast<Isolate*>(ptr);
     Locker locker(iso);
     Isolate::Scope isolate_scope(iso);
     HandleScope handle_scope(iso);
   
     iso->SetCaptureStackTraceForUncaughtExceptions(true);
+
+    v8::Local<v8::ObjectTemplate> global = v8::ObjectTemplate::New(iso);
+    global->Set(v8::String::NewFromUtf8(iso, "print", NewStringType::kNormal).ToLocalChecked(),
```
rogchap, line 164: Would it not make more sense to use a familiar JavaScript API; for example `console.log` `console.error` etc

v8go.cc
```diff
@@ -134,16 +134,38 @@ IsolateHStatistics IsolationGetHeapStatistics(IsolatePtr ptr) {
...
+// Print provides a basic interface to print to stdout.
+static void Print(const FunctionCallbackInfo<Value>& args) {
+  bool first = true;
+  auto iso = args.GetIsolate();
+  for (int i = 0; i < args.Length(); i++) {
+    HandleScope handle_scope(iso);
+    if (first) {
+      first = false;
+    } else {
+      printf(" ");
```
rogchap, line 146: `printf` and `println` builtins are  not designed for production code:

> Print is useful for bootstrapping and debugging; it is not guaranteed to stay in the language. 

Would be better if we could find an API that uses an `io.Writer` so that you can set to `os.Stdout` or maybe `os.Stderr` but also any other file you want to "print" too.
https://tip.golang.org/pkg/builtin/#print

v8go.cc
```diff
@@ -134,16 +134,38 @@ IsolateHStatistics IsolationGetHeapStatistics(IsolatePtr ptr) {
...
+    }
+    String::Utf8Value value(iso, args[i]);
+    printf("%s", *value);
+  }
+  printf("\n");
+  fflush(stdout);
```
rogchap, line 152: What is this function? I can't see it defined. To flush stdout you would normally do something like:
```
f := bufio.NewWriter(os.Stdout)
defer f.Flush()
f.Write(b)
```

v8go.cc
rogchap: Need to think about this API in a bit more depth. If we are looking to provide some sort of interop then maybe we need a `v8go` object.... eg: `v8go.print()`; That said I'm not 100% convinced in having "custom" code withing your Javascript code.

rogchap (changes requested):

rogchap:
@tmc Over the holidays I've been thinking about the best way to move forward with this.

I think the best thing to do for this project is provide an interface for callers to create there own `ObjectTemplate`, so that they would be able to create their own `print` function or any other type of global function.

Quite a bit more work, but infinity more powerful. Thoughts?

tmc:
Yes that makes for a more flexible approach.

zwang:
This will be great, we keep the API small but very powerful.

> I think the best thing to do for this project is provide an interface for callers to create there own ObjectTemplate, so that they would be able to create their own print function or any other type of global function.

PR #42, opened 2020-11-29
tmc windows build

tmc:


rogchap:
Thanks @tmc for trying to get this working. I think dynamic linking might be the best compromise for Windows (and maybe other archs); So I think we should use #46 as a path forward for Windows. Thoughts?

tmc:
Happy to see any path forward working!

tmc:
Closing this in favor of #46

PR #50, opened 2020-12-19
Value methods to assert value kind

rogchap:
Expanding the API for the `Value` struct to include checks for the value returned by a run script.
This will pave the way forward for more API's to interact between Go and Javascript.


tmc (commented):
Love this direction!

zwass:
Nice! This looks to be a precursor to being able to extract the values (beyond the existing string method)?

rogchap:
That's right; once I have this PR done, I'll add the `ToXXXX` methods (as a seperate PR to keep it easier to review)

rogchap:
Hey @tmc Would you mind giving this a sanity check?

.github/workflows/test.yml
```diff
@@ -1,6 +1,11 @@
 name: CI
 
-on: [push, pull_request, workflow_dispatch]
+on:
```
tmc, line 3: did you mean to change this to only run on master? it seems like this should run on any push.

nit: whitespace at EOL

v8go.cc
```diff
@@ -200,26 +200,296 @@ void ContextDispose(ContextPtr ptr) {
...
+    Isolate* iso = ctx->iso; \
+    Locker locker(iso); \
+    Isolate::Scope isolate_scope(iso); \
+    HandleScope handle_scope(iso); \
+    Context::Scope context_scope(ctx->ptr.Get(iso)); \
+    Local<Value> value = val->ptr.Get(iso);
+
 void ValueDispose(ValuePtr ptr) {
-  delete static_cast<m_value*>(ptr);
+    delete static_cast<m_value*>(ptr);
```
tmc, line 214: would love to get clangtidy/otherwise in here to standardize the non-go code formatting as well.

tmc (commented):
This looks reasonable to me, added a few small comments.

would you mind populating the PR description for posterity's sake?

.github/workflows/test.yml
rogchap: This is a quirk of GitHub actions. I previously had just `push` but that does not work for pushes to forks, so I added `pull_request` but now if I (and only people that can push directly to this repo) submit a pull request the tests get run twice: Once for `push` and again for `pull_request`.
There should be no other branches other than master that should need to run the test unless you are creating a pull request. If you really need to then that's what `workflow_dispatch` will allow, but is a manual process.

v8go.cc
rogchap: Agree; will save that for a separate PR; I did format the whole file, but I thought that was too much of a change to add to this PR, so I reverted that. (missed this line though).

rogchap: #51

tmc (approved):
LGTM!

PR #59, opened 2020-12-22
Add (Basic) getter methods for the v8 value struct

rogchap:
This will enable getting of richer data types from the v8 value struct than just currently `string`

All the basic `ToXXX` methods have been implemented as defined here: 
https://v8.github.io/api/head/classv8_1_1Value.html#a838af7d16f0ce87b4411beb178383db4

However `ToBigInt()` and `ToObject()` have NOT yet been implemented. The reason is that `Object` represents everything that is not a primitive type: eg. Functions, Objects, Promises, Dates, Regex etc. Implementing all of the `Object` type would make this PR too large, and take quite a bit of time to implement.

Was hoping to include this PR in the next release, which already has a number of features people are waiting for to be released.

As such, I think it's best to un-export these methods (to not break the public API), get this PR merged, do a release and continue work on `Object` (and `BigInt`) in later PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Value methods for checking value kind (is string, number, array etc)
 - C formatting via `clang-format` to aid future development
 - Support of vendoring with `go mod vendor`
+- Value methods to convert to primitive data types
 
 ### Changed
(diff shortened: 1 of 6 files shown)


tmc (approved):
This looks great! We should consider issues for the bigInt and object implementations. I like the fmt.Formatter support addition.

rogchap:
@tmc Thanks. I've added those to issues as suggested. ☝️

PR #63, opened 2021-01-15
Support JS BigInt value to big.Int Go type

### LINKED ISSUE #62: Support for the BigInt value type
`BigInt` is a primitive within V8 that is currently missing from the Value struct within v8go (as implemented in #59).
Docs: https://v8.github.io/api/head/classv8_1_1BigInt.html

Would be good to see if we could map back to the `big.Int` within the `math/big` stdlib pkg: https://golang.org/pkg/math/big/ 
Or provide our own `v8go.BigInt` struct if that is not possible.

rogchap:
resolves #62

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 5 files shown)


tmc (approved):
Looks reasonable to me. There might be additional test case coverage we'd want but this looks great!

PR #64, opened 2021-01-17
Support creating a global ObjectTemplate into a context

rogchap:
Provides creating a global object that can be initiated within each Context eg:
```go
iso, _ := v8go.NewIsolate()
obj, _ := v8go.NewObjectTemplate(iso)
obj.Set("version", "v1.0.0", v8go.ReadOnly)
ctx, _ := v8go.NewContext(iso, obj)
val, _ := ctx.RunScript("version", "main.js")
fmt.Println(val)
// Output:
// v1.0.0
```
The same `ObjectTemplate` can be used in many Contexts as long as they are part of the same Isolate (VM).

This PR also adds the ability to create primitive values that can be properties on the global object.

This is the prerequisite to `FunctionTemplate`s for the global context.

As a side-effect of this PR. The `NewContext()` API has changed to optionally accept a Isolate and a Global Template, this makes for a nice API as you see above, but also allows for this:
```go
ctx, _ := v8go.NewContext()    // previously you would do v8go.NewContext(nil)
```

This is fully backwards compatible.

I've not updated the README.md to not confuse readers, until this is released into a new version.

### DIFF
--- a/context_test.go
+++ b/context_test.go
@@ -74,3 +74,39 @@ func BenchmarkContext(b *testing.B) {
 		ctx.Close()
 	}
 }
+
+func ExampleContext() {
+	ctx, _ := v8go.NewContext()
+	ctx.RunScript("const add = (a, b) => a + b", "math.js")
+	ctx.RunScript("const result = add(3, 4)", "main.js")
+	val, _ := ctx.RunScript("result", "value.js")
+	fmt.Println(val)
+	// Output:
+	// 7
+}
+
+func ExampleContext_isolate() {
+	iso, _ := v8go.NewIsolate()
+	ctx1, _ := v8go.NewContext(iso)
+	ctx1.RunScript("const foo = 'bar'", "context_one.js")
+	val, _ := ctx1.RunScript("foo", "foo.js")
+	fmt.Println(val)
+
+	ctx2, _ := v8go.NewContext(iso)
+	_, err := ctx2.RunScript("foo", "context_two.js")
+	fmt.Println(err)
+	// Output:
+	// bar
+	// ReferenceError: foo is not defined
+}
+
+func ExampleContext_globalTemplate() {
+	iso, _ := v8go.NewIsolate()
+	obj, _ := v8go.NewObjectTemplate(iso)
+	obj.Set("version", "v1.0.0", 0)
+	ctx, _ := v8go.NewContext(iso, obj)
+	val, _ := ctx.RunScript("version", "main.js")
+	fmt.Println(val)
+	// Output:
+	// v1.0.0
+}
--- a/context_test.go
+++ b/context_test.go
@@ -103,7 +103,7 @@ func ExampleContext_isolate() {
 func ExampleContext_globalTemplate() {
 	iso, _ := v8go.NewIsolate()
 	obj, _ := v8go.NewObjectTemplate(iso)
-	obj.Set("version", "v1.0.0", 0)
+	obj.Set("version", "v1.0.0")
 	ctx, _ := v8go.NewContext(iso, obj)
 	val, _ := ctx.RunScript("version", "main.js")
 	fmt.Println(val)
--- a/object_template.go
+++ b/object_template.go
@@ -33,6 +33,7 @@ type ObjectTemplate struct {
 }
 
 // NewObjectTemplate creates a new ObjectTemplate.
+// The *ObjectTemplate can be used as a v8go.ContextOption to create a global object in a Context.
 func NewObjectTemplate(iso *Isolate) (*ObjectTemplate, error) {
 	if iso == nil {
 		return nil, errors.New("object_template: failed to create new ObjectTemplate: isolate cannot be <nil>")
--- a/object_template.go
+++ b/object_template.go
@@ -1,0 +1,74 @@
+package v8go
+
+// #include <stdlib.h>
+// #include "v8go.h"
+import "C"
+import (
+	"errors"
+	"fmt"
+	"runtime"
+	"unsafe"
+)
+
+// PropertyAttribute are the attributes for a property
+type PropertyAttribute uint8
+
+const (
+	// None.
+	None PropertyAttribute = 0
+	// ReadOnly, ie. not writable.
+	ReadOnly = 1 << iota
+	// DontEnum, ie. not enumerable.
+	DontEnum
+	// DontDelete, ie. not configurable.
+	DontDelete
+)
+
+// ObjectTemplate is used to create objects at runtime.
+// Properties added to an ObjectTemplate are added to each object created from the ObjectTemplate.
+type ObjectTemplate struct {
+	ptr C.ObjectTemplatePtr
+}
+
+// NewObjectTemplate creates a new ObjectTemplate.
+func NewObjectTemplate(iso *Isolate) (*ObjectTemplate, error) {
+	if iso == nil {
+		return nil, errors.New("object_template: failed to create new ObjectTemplate: isolate cannot be <nil>")
+	}
+	ob := &ObjectTemplate{
+		ptr: C.NewObjectTemplate(iso.ptr),
+	}
+	runtime.SetFinalizer(ob, (*ObjectTemplate).finalizer)
+	return ob, nil
+}
+
+// Set adds a property to each instance created by this template.
+// The property must be defined either as a primitive value, or a template.
+func (o *ObjectTemplate) Set(name string, val interface{}, attributes PropertyAttribute) error {
+	cname := C.CString(name)
+	defer C.free(unsafe.Pointer(cname))
+
+	switch v := val.(type) {
+	case *ObjectTemplate:
+		panic("not implemented")
+	case *Value:
+		if v.IsObject() || v.IsExternal() {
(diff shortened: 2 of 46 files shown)


object_template.go
```diff
@@ -0,0 +1,86 @@
...
+import "C"
+import (
+	"errors"
+	"fmt"
+	"math/big"
+	"runtime"
+	"unsafe"
+)
+
+// PropertyAttribute are the attributes for a property
```
tmc, line 14: nit: no period.

object_template.go
```diff
@@ -0,0 +1,86 @@
...
+type ObjectTemplate struct {
+	ptr C.ObjectTemplatePtr
+	iso *Isolate
+}
+
+// NewObjectTemplate creates a new ObjectTemplate.
+// The *ObjectTemplate can be used as a v8go.ContextOption to create a global object in a Context.
+func NewObjectTemplate(iso *Isolate) (*ObjectTemplate, error) {
+	if iso == nil {
+		return nil, errors.New("object_template: failed to create new ObjectTemplate: isolate cannot be <nil>")
```
tmc, line 39: kind of odd to prefix with object_template here -- consider omitting or using v8go. You could also just simpify dramatically via:
"isolate cannot be nil"

object_template.go
```diff
@@ -0,0 +1,86 @@
...
+		C.ObjectTemplateSetValue(o.ptr, cname, newVal.ptr, C.int(attributes))
+	case *ObjectTemplate:
+		C.ObjectTemplateSetObjectTemplate(o.ptr, cname, v.ptr, C.int(attributes))
+	case *Value:
+		if v.IsObject() || v.IsExternal() {
+			return errors.New("object_template: unsupported property: value type must be a primitive or use a template")
+		}
+		C.ObjectTemplateSetValue(o.ptr, cname, v.ptr, C.int(attributes))
+	default:
+		return fmt.Errorf("object_template: unsupported property type `%T`, must be one of string, int32, uint32, int64, uint64, float64, *big.Int, *v8go.Value or *v8go.ObjectTemplate", v)
```
tmc, line 72: similarly --  might be more appropriate to omit the prefix or make it v8go?

object_template.go
```diff
@@ -0,0 +1,86 @@
...
+		iso: iso,
+	}
+	runtime.SetFinalizer(ob, (*ObjectTemplate).finalizer)
+	return ob, nil
+}
+
+// Set adds a property to each instance created by this template.
+// The property must be defined either as a primitive value, or a template.
+// If the value passed is a Go supported primitive (string, int32, uint32, int64, uint64, float64, big.Int)
+// then a value will be created an set as the value property.
```
tmc, line 52: an -> and

object_template.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	return nil
+}
+
+func (o *ObjectTemplate) apply(opts *contextOptions) {
+	opts.gTmpl = o
+}
+
+func (o *ObjectTemplate) finalizer() {
```
tmc, line 82: for transparency: I don't totally grok this -- maybe add a comment (or two?)

value.go
```diff
@@ -14,6 +14,87 @@ import (
...
+// NewValue will create a primitive value. Supported values types to create are:
+//   string -> V8::String
+//   int32 -> V8::Integer
+// 	 uint32 -> V8::Integer
+//	 bool -> V8::Boolean
+//   int64 -> V8::BigInt
+//   uint64 -> V8::BigInt
+//	 bool -> V8::Boolean
+// 	 *big.Int -> V8::BigInt
+func NewValue(iso *Isolate, val interface{}) (*Value, error) {
```
tmc, line 29: this should probably include a nil check for iso

tmc (commented):
This is looking great -- mostly minor comments.

context_test.go
```diff
@@ -74,3 +74,39 @@ func BenchmarkContext(b *testing.B) {
...
+	fmt.Println(err)
+	// Output:
+	// bar
+	// ReferenceError: foo is not defined
+}
+
+func ExampleContext_globalTemplate() {
+	iso, _ := v8go.NewIsolate()
+	obj, _ := v8go.NewObjectTemplate(iso)
+	obj.Set("version", "v1.0.0", 0)
```
tmc, line 106: this example should use the constant instead of the magic value 0

tmc:
mentioned in the diff inline but in the PR description as well, the Set call's attribute just being passed as "0" is a bit blah.

perhaps we should consider a different api to set the property attributes so that could be omitted in the normal case?

object_template.go
```diff
@@ -0,0 +1,86 @@
...
+	"runtime"
+	"unsafe"
+)
+
+// PropertyAttribute are the attributes for a property
+type PropertyAttribute uint8
+
+const (
+	// None.
+	None PropertyAttribute = 0
```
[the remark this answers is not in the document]
rogchap: That would be incorrect, because that would mean `ReadOnly` would result in `1 << 1` and it needs to be `1 << 0`
As per the doc:
> enum | PropertyAttribute { None = 0,  ReadOnly = 1 << 0,  DontEnum = 1 << 1,  DontDelete = 1 << 2  }
https://v8.github.io/api/head/namespacev8.html

object_template.go
rogchap: Fair enough. Should really only prefix with the the package name ie. `v8go` but I was following a previous pattern (that I wrongly did for the value struct). Will amend all error messages for consistency.

object_template.go
rogchap: This is an internal function that is not exposed to the caller.
This is part of the management of the GC from Go and cleaning up the data in C.
Recommend reading: https://golang.org/pkg/runtime/#SetFinalizer

value.go
rogchap: Can do. Would fail pretty quickly without the VM running 😉 
Also noticed the poor formatting of the doc comment, will fix that too.

rogchap:
> Set call's attribute just being passed as "0" is a bit blah.
> perhaps we should consider a different api to set the property attributes so that could be omitted in the normal case?

That's me just being lazy. I'll change the examples to use the enum/const.
I'm open to suggestions on the API, however this is the API that is provided by the C++ API, and I want to keep as true to that as possible. 

For ObjectTemplate's it's more likely that you want to set `ReadOnly` rather than `None` anyway.

object_template.go
rogchap: Hahah. The external docs are even less informative: https://v8.github.io/api/head/namespacev8.html#a05f25f935e108a1ea2d150e274602b87

tmc:
That's fair, I think attempting to match the c++ makes sense (and I know it's stated goal).

object_template.go
tmc: yeah I just meant that an internal comment might be nice for readers/contributors

object_template.go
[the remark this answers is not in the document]
tmc: Ah sorry, you're right there.

This does need the type on the iota line else these end up as untyped constants.

object_template.go
rogchap: hmmm... I'm not a fan of over commenting the code as the code (should) speaks for itself; and in my opinion it's fairly clear what this function does by calling `runtime.SetFinalizer`.

rogchap:
> It's just too bad that something as succinct as x.Set("foo", "bar") isn't possible

The goals to state o match the C++ API, but still as idiomatic Go as possible (our get out of jail free card 😉 ).
What we could do is change this API to a variadic function, similar to what we are doing for Context.

This would make the property attribute optional and default to `None`.

Ps. the PropertyAttribute is also used for normal Objects (not template ones); so would have to do the same there too.

What do you think @tmc ?

object_template.go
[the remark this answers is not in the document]
rogchap: good spot. Without it would end up as an int.

rogchap:
So thinking about this a bit more... rather than a varadic function with "SetOptions"... what about just doing a varadic function of PropertyAttributes, because they are flags that can be added together?

This is better explained via examples:

### Current API

```go
func (o *ObjectTemplate) Set(name string, val interface{}, attributes PropertyAttribute) error 

// the caller would do:
obj.Set("foo", "bar", v8go.ReadOnly)
// or for multiple:
obj.Set("foo", "bar, v8go.ReadOnly | v8go.DontEnum)
```

### Proposed API (v2)

```go
func (o *ObjectTemplate) Set(name string, val interface{}, attributes ...PropertyAttribute) error 

// previous examples would still be valid
obj.Set("foo", "bar", v8go.ReadOnly)
obj.Set("foo", "bar, v8go.ReadOnly | v8go.DontEnum)

// but would also allow for this
obj.Set("foo", "bar")
obj.Set("foo", "bar, v8go.ReadOnly, v8go.DontEnum)
```

In the last example the slice of attributes would be converted to `OR` ie.
```go
obj.Set("foo", "bar, v8go.ReadOnly, v8go.DontEnum)
// is equal to 
obj.Set("foo", "bar, v8go.ReadOnly | v8go.DontEnum)

tmc (approved):
LGTM!

Issue #61, opened 2021-01-14
Support for the Object value type


rogchap:
As documented here: https://v8.github.io/api/head/classv8_1_1Object.html

The Object class represents the non-primitive data types. ie (Objects, Arrays, Promises, Dates, Regex etc)
Primitive data types where implemented in #59 

The main "entry" to the Object should be from the `Value` struct. ie. `val.Object()`. 

This may require a multi-PR as the Object class is quite large, and breaking it down will make implementation and reviews easier.

PR #68, opened 2021-01-26
Function Templates with callback functions in Go

rogchap:
Basic `FunctionTemplate` support, which allows global functions that callback to Go functions:
```go
iso, _ := v8go.NewIsolate()
global, _ := v8go.NewObjectTemplate(iso)
printfn, _ := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *v8go.Value {
	fmt.Printf("%+v\n", info.Args())
	return nil
})
global.Set("print", printfn, v8go.ReadOnly)
ctx, _ := v8go.NewContext(iso, global)
ctx.RunScript("print('foo', 'bar', 0, 1)", "")
// Output:
// [foo bar 0 1]
```

Due to the complexities of Go -> C -> Go there are two registries:

1) Callback registry attached to the Isolate
2) A global registry for the Context

To make sure that created Contexts can be GC'd we have a register/deregister surrounding script execution; other wise a `map[int]*Context` would never GC

### DIFF
--- a/function_template.go
+++ b/function_template.go
@@ -55,15 +55,15 @@ func NewFunctionTemplate(iso *Isolate, callback FunctionCallback) (*FunctionTemp
 }
 
 //export goFunctionCallback
-func goFunctionCallback(ctxref int, cbref int, args *C.ValuePtr, args_count int) C.ValuePtr {
+func goFunctionCallback(ctxref int, cbref int, args *C.ValuePtr, argsCount int) C.ValuePtr {
 	ctx := getContext(ctxref)
 
 	info := &FunctionCallbackInfo{
 		ctx:  ctx,
-		args: make([]*Value, args_count),
+		args: make([]*Value, argsCount),
 	}
 
-	argv := (*[1 << 30]C.ValuePtr)(unsafe.Pointer(args))[:args_count:args_count]
+	argv := (*[1 << 30]C.ValuePtr)(unsafe.Pointer(args))[:argsCount:argsCount]
 	for i, v := range argv {
 		val := &Value{ptr: v}
 		runtime.SetFinalizer(val, (*Value).finalizer)
--- a/function_template.go
+++ b/function_template.go
@@ -1,0 +1,59 @@
+package v8go
+
+// #include <stdlib.h>
+// #include "v8go.h"
+import "C"
+import (
+	"errors"
+	"runtime"
+	"unsafe"
+)
+
+type FunctionCallback func(info *FunctionCallbackInfo) *Value
+
+type FunctionCallbackInfo struct {
+	args []*Value
+}
+
+func (i *FunctionCallbackInfo) Args() []*Value {
+	return i.args
+}
+
+type FunctionTemplate struct {
+	*template
+}
+
+func NewFunctionTemplate(iso *Isolate, callback FunctionCallback) (*FunctionTemplate, error) {
+	if iso == nil {
+		return nil, errors.New("v8go: failed to create new FunctionTemplate: Isolate cannot be <nil>")
+	}
+
+	tmpl := &template{
+		ptr: C.NewFunctionTemplate(iso.ptr, unsafe.Pointer(&callback)),
+		iso: iso,
+	}
+	runtime.SetFinalizer(tmpl, (*template).finalizer)
+	return &FunctionTemplate{tmpl}, nil
+}
+
+//export goFunctionCallback
+func goFunctionCallback(callback unsafe.Pointer, args *C.ValuePtr, args_count int) {
+	callbackFunc := *(*FunctionCallback)(callback)
+
+	//TODO: This will need access to the Context to be able to create return values
+	info := &FunctionCallbackInfo{
+		args: make([]*Value, args_count),
+	}
+
+	argv := (*[1 << 30]C.ValuePtr)(unsafe.Pointer(args))[:args_count:args_count]
+	for i, v := range argv {
+		//TODO(rogchap): We must pass the current context and add this to the value struct
+		val := &Value{ptr: v}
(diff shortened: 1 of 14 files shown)


v8go.cc
```diff
@@ -148,79 +165,152 @@ IsolateHStatistics IsolationGetHeapStatistics(IsolatePtr ptr) {
...
-    global_template = ob->ptr.Get(iso);
+    m_template* ob = static_cast<m_template*>(global_template_ptr);
+    global_template = ob->ptr.Get(iso).As<ObjectTemplate>();
   } else {
     global_template = ObjectTemplate::New(iso);
   }
 
-  iso->SetCaptureStackTraceForUncaughtExceptions(true);
+  Local<Context> local_ctx = Context::New(iso, nullptr, global_template);
+  local_ctx->SetEmbedderData(1, Integer::New(iso, ref));
```
tmc, line 294: I had to dig a bit to understand how this was used -- could you add some comments in here?

v8go.cc
```diff
@@ -148,79 +165,152 @@ IsolateHStatistics IsolationGetHeapStatistics(IsolatePtr ptr) {
...
   ot->iso = iso;
   ot->ptr.Reset(iso, ObjectTemplate::New(iso));
-  return static_cast<ObjectTemplatePtr>(ot);
+  return static_cast<TemplatePtr>(ot);
 }
+/********** FunctionTemplate **********/
 
-void ObjectTemplateDispose(ObjectTemplatePtr ptr) {
-  delete static_cast<m_object_template*>(ptr);
+static void FunctionTemplateCallback(const FunctionCallbackInfo<Value>& info) {
```
tmc, line 221: this is a little bit tricky and would love to see some comments in here as well!

tmc (approved):
There's a lot going on here and I can't say I grok every line but I'm going to approve as the tests look good and I don't see glaring errors!

isolate.go
```diff
@@ -72,17 +81,40 @@ func (i *Isolate) GetHeapStatistics() HeapStatistics {
...
-	C.IsolateDispose(i.ptr)
-	i.ptr = nil
-	runtime.SetFinalizer(i, nil)
+// Dispose will dispose the Isolate VM; subsequent calls will panic.
+func (i *Isolate) Dispose() {
+	i.finalizer()
 }
 
-// Close will dispose the Isolate VM; subsequent calls will panic
+// Close is Deprecated. Use `iso.Dispose()`.
```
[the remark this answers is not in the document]
rogchap: Annoyingly the semi-standard "Deprecated: " prefix violates `golint`... that said `golint` has also been deprecated! 🤷‍♂️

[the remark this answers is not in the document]
tmc: I've moved to use staticcheck and go-critic, FWIW.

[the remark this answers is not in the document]
rogchap: Weirdly I could not get gocritic to "error/warn" on the doc-comment even though it's listed as a check 🤷‍♂️ 
Oh well.

v8go.cc
rogchap: Added some comments, and actually simplified the data object. Hope it makes a little more sense now? (as much as it can with V8 C++ code)

function_template.go
```diff
@@ -0,0 +1,78 @@
...
+
+// FunctionTemplate is used to create functions at runtime.
+// There can only be one function created from a FunctionTemplate in a context.
+// The lifetime of the created function is equal to the lifetime of the context.
+type FunctionTemplate struct {
+	*template
+}
+
+// NewFunctionTemplate creates a FunctionTemplate for a given callback.
+func NewFunctionTemplate(iso *Isolate, callback FunctionCallback) (*FunctionTemplate, error) {
```
choonkeat, line 39: how does `vm.TerminateExecution` interact with callback of `v8go.NewFunctionTemplate`?

would it be possible to include `context.Context` as first argument of callback?

rogchap: Once the VM is terminated, you will no longer be able to fire the callback. These callbacks are connected to the isolate, so once the Isolate is GC'd so will be the callback functions (unless attached to some other struct etc)

You can get the current context via `info.Context()`

rogchap: `context.Context` will get confusing with `v8go.Context` (as I just did)
The V8 API does not have such a concept as the Go Context; if we introduce the Go Context, this would have to exist throughout the API, which would likely be a breaking change.

rogchap: I had a look at you code (https://github.com/choonkeat/try-v8go/blob/main/main.go)
I see what you are trying to do... you want to be able to check if the Go Context is `Done` within the callback?
Maybe we could attach a context in a way that is non-breaking ie. something like `info.GoContext()` ??

rogchap: @choonkeat Would you be able to submit this Issue as a feature request? It's maybe something we could look to supporting.

choonkeat: > context.Context will get confusing with v8go.Context (as I just did)

ya :-P

> Would you be able to submit this Issue as a feature request?

it was more of a comment, since idiomatic Go callback function usually comes with `context.Context`. _But_ if it doesn't fit v8, I don't mind passing my own `context.Context` via closure

---

i'm still trying to wrap my head around how things "should" work. e.g. implementing `fetch` i'd need to return a Promise i think and thus will be tracking https://github.com/rogchap/v8go/issues/61 closely

rogchap: Yes, to implement true `fetch` you need to return the Promise; A poor man's version is here: https://github.com/rogchap/v8go/blob/master/function_template_test.go#L52

rogchap: FYI. this is better supported now that we have #76

PR #70, opened 2021-01-31
JSON parse and stringify API

rogchap:
Simple JSON API as described here: https://v8.github.io/api/head/classv8_1_1JSON.html

### DIFF
--- a/context.go
+++ b/context.go
@@ -4,14 +4,11 @@ package v8go
 // #include "v8go.h"
 import "C"
 import (
-	"errors"
 	"fmt"
 	"runtime"
 	"sync"
 	"unsafe"
 )
-
-var noContextErr = errors.New("v8go: valid Context is required")
 
 // Due to the limitations of passing pointers to C from Go we need to create
 // a registry so that we can lookup the Context from any given callback from V8.
--- a/context.go
+++ b/context.go
@@ -4,11 +4,14 @@ package v8go
 // #include "v8go.h"
 import "C"
 import (
+	"errors"
 	"fmt"
 	"runtime"
 	"sync"
 	"unsafe"
 )
+
+var noContextErr = errors.New("v8go: valid Context is required")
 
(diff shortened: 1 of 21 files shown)


context.go
```diff
@@ -77,6 +77,23 @@ func (c *Context) RunScript(source string, origin string) (*Value, error) {
...
+func (c *Context) JSONParse(str string) (*Value, error) {
+	cstr := C.CString(str)
+	defer C.free(unsafe.Pointer(cstr))
+
+	rtn := C.JSONParse(c.ptr, cstr)
+	return getValue(c, rtn), getError(rtn)
+}
+
+// JSONStringify tries to stringify the JSON-serializable object value and returns it as string.
+func (c *Context) JSONStringify(val *Value) string {
```
tmc, line 91: sidebar: Would be great to get some benchmarks into CI and report on # of allocs.

tmc (approved):
LGTM!

kuoruan:
I think the API like `ctx.JSON.Stringfy(...)` and `ctx.JSON.Parse(...)` is better.

In the future, we can add `ctx.Math.Min(...)` etc...

rogchap:
Thanks for the feedback @kuoruan 

The API you have suggested requires an extra allocation, with a redundant structure. I don't believe that this pattern is idiomatic Go (I can't think of any examples in the Stdlib). FYI: There is no `Math` API in V8; you would do this in either Go or JS directly.

The only alternative design would to make the API functions rather than methods; which is technically closer to the C++ API ie:
```go
func JSONParse(ctx *Context, str string) (*Value, error)

func JSONStringify(ctx *Context, val *Value) string
``` 
We could also place them in there own `json` pacakge: ie:
```go
import "rogchap.com/v8go/json"

val := json.Parse(ctx, str)
jsonStr := json.Stringify(ctx, val)
```
But seems unnecessary to create a new package for these 2 functions

tmc (approved):
LGTM

PR #73, opened 2021-02-04
BSD-style licence

rogchap:
Updates the LICENSE for v8go to be **3 Clause BSD** from MIT
BSD has a minute difference to MIT and are basically "the same" (but I'm not a lawyer so... 🤷‍♂️)

This keeps the license in-line with the V8 library (and Go itself) that also has a the 3 Clause BSD license.


zwang (approved):
Looks good to me!

zwang (approved):

PR #74, opened 2021-02-04
Extend Value into the Object type

rogchap:
Basic casting of a Value to an Object (`val.AsObject()`) as well as converting Value to an Object (`val.Object()`)

Only covers a small subset of the Object API as described here: https://v8.github.io/api/head/classv8_1_1Object.html

Also includes converting an `ObjectTemplate` to an `Object` with a given Context, and retrieving the Global Object from a Context.

This allows for things like this at runtime: 
```go
iso, _ := v8go.NewIsolate()
obj, _ := v8go.NewObjectTemplate(iso)
ctx, _ := v8go.NewContext(iso, obj)
global := ctx.Global()
console, _ := v8go.NewObjectTemplate(iso)
logfn, _ := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *v8go.Value {
	fmt.Println(info.Args()[0])
	return nil
})
console.Set("log", logfn)
consoleObj, _ := console.NewInstance(ctx)
global.Set("console", consoleObj)
ctx.RunScript("console.log('foo')", "")
// Output:
// foo
```

### DIFF
--- a/json.go
+++ b/json.go
@@ -22,8 +22,8 @@ func JSONParse(ctx *Context, str string) (*Value, error) {
 }
 
 // JSONStringify tries to stringify the JSON-serializable object value and returns it as string.
-func JSONStringify(ctx *Context, val *Value) (string, error) {
-	if val == nil {
+func JSONStringify(ctx *Context, val Valuer) (string, error) {
+	if val.value() == nil {
 		return "", errors.New("v8go: Value is required")
 	}
 	// If a nil context is passed we'll use the context/isolate that created the value.
@@ -32,7 +32,7 @@ func JSONStringify(ctx *Context, val *Value) (string, error) {
 		ctxPtr = ctx.ptr
 	}
 
-	str := C.JSONStringify(ctxPtr, val.ptr)
+	str := C.JSONStringify(ctxPtr, val.value().ptr)
 	defer C.free(unsafe.Pointer(str))
 	return C.GoString(str), nil
 }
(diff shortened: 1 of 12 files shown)


zwang:
Nice, will this changes to allow getting any object and also override to set any object? Probably except Global Object?

rogchap:
Yes, you should be able to get and set any Object in the tree and the change would be reflected in the VM.
You can change the Global Object too; but do so at your own peril, as you could break the VM.

json.go
```diff
@@ -26,8 +26,8 @@ func JSONParse(ctx *Context, str string) (*Value, error) {
 }
 
 // JSONStringify tries to stringify the JSON-serializable object value and returns it as string.
-func JSONStringify(ctx *Context, val *Value) (string, error) {
-	if val == nil {
+func JSONStringify(ctx *Context, val Valuer) (string, error) {
+	if val.value() == nil {
```
tmc, line 30: Should we also check for nil val?

object_template_test.go
```diff
@@ -117,3 +117,20 @@ func TestGlobalObjectTemplate(t *testing.T) {
 		})
 	}
 }
+
+func TestObjectTemplateNewInstace(t *testing.T) {
```
tmc, line 121: Nit: spelling on Instance

tmc (approved):
Only minor comments. LGTM

PR #76, opened 2021-02-11
Promise resolver and promise result

### LINKED ISSUE #69: [feature] How to get the Promise result value?
```
ctx, _ := v8go.NewContext(nil)

val, err := ctx.RunScript(`new Promise((resolve) => setTimeout(() => resolve(1), 1000)).then(console.log)`, "value.js")
if err != nil {
  panic(err)
}

fmt.Printf("result: %s", val.String())
```

```
result: [object Promise]
```

I want to get the Promise result, for the example code is `1`

rogchap:
Enables the (basic) Promise API. eg:
```go
iso, _ := v8go.NewIsolate()
global, _ := v8go.NewObjectTemplate(iso)

fetchfn, _ := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *v8go.Value {
	args := info.Args()
	url := args[0].String()
	resolver, _ := v8go.NewPromiseResolver(info.Context())

	go func() {
		res, _ := http.Get(url)
		body, _ := ioutil.ReadAll(res.Body)
		val, _ := v8go.NewValue(iso, string(body))
		resolver.Resolve(val)
	}()
	return resolver.GetPromise().Value
})
global.Set("fetch", fetchfn, v8go.ReadOnly)

ctx, _ := v8go.NewContext(iso, global)
val, _ := ctx.RunScript("fetch('https://rogchap.com/v8go')", "")
prom, _ := val.AsPromise()

// wait for the promise to resolve
for prom.State() == v8go.Pending {
	continue
}
fmt.Printf("%s\n", strings.Split(prom.Result().String(), "\n")[0])
// Output:
// <!DOCTYPE html>
```
resolves #69


kuoruan:
Dose `exception.go` required when reject a promise?
Can I just reject a string?

```
e, _ := v8go.NewValue(iso, "any errors")
resolver.Reject(e)
```

rogchap:
You should be able to reject with any "Value" which includes a string, but it's most common to use an Error/Exception

rogchap:
Just a quick update:

Sorry for taking so long in getting this PR finished; it's almost done.

Annoyingly, I'm in the process of moving house, so packing boxes has been taking most of my spare time 😞

Should be fully moved in by this time next week, and should be able to continue my work on v8go.

jensneuse:
I'm not experienced with v8 but like to ask a question.
Looking at this code:
```go
for prom.State() == v8go.Pending {
	continue
}
```
It seems, this is constantly calling the cgo bridge, doesn't it?
Will this affect runtime performance?
Would it make sense to blocking read on a channel until the v8 calls a resolved/rejected callback?

rogchap:
@jensneuse Thanks for the suggestion. Hopefully this helps answer your questions:

> It seems, this is constantly calling the cgo bridge, doesn't it?

It does 😓 

> Will this affect runtime performance?

Not really. The main issues around performance arise from transferring large amounts of data between C and Go and having to copy that data into memory on the Go side to make sure it does not get freed on the C side before you use it on the Go side. This func is returning an enum aka integer by value so you shouldn't notice any overhead in doing this in Go over doing the same thing in C++.

> Would it make sense to blocking read on a channel until the v8 calls a resolved/rejected callback?

It's a nice idea, but one of the goals of the project is stay as close to the V8 API as possible; it also has a lot larger overhead as callbacks are not an easy thing to manage between Go and C (Have a look at the FunctionTemplate code to get an idea of the complexity)

rogchap:
I'm having a hard time with getting Error Exceptions working. This simple line is causing a segfault:
```cpp
Exception::Error(String::Empty(iso));
```
I've verified that the `Isolate*` is valid as I can create other values, just not Exception value. 🤷‍♂️ 

It's not critical for this PR to have the Exception/Error objects, but I thought it would be nice for the Promise rejections.
As noted above, you can use any Value as the value to the Reject function. 

I'm tempted to exclude the `exception.go` stuff (just make it private for now) and revisit in another PR.

tmc:
> I'm tempted to exclude the `exception.go` stuff (just make it private for now) and revisit in another PR.

It seems like that can come in a follow-on commit.

promise.go
```diff
@@ -0,0 +1,87 @@
...
+// Promise is the JavaScript promise object defined in ES6
+type Promise struct {
+	*Object
+}
+
+// MewPromiseResolver creates a new Promise resolver for the given context.
+// The associated Promise will be in a Pending state.
+func NewPromiseResolver(ctx *Context) (*PromiseResolver, error) {
+	if ctx == nil {
+		return nil, errors.New("v8go: Context is required")
```
tmc, line 40: nite: lowercase?

tmc (approved):
LGTM, some small comments.

zwang (approved):

kuoruan:
Hi @rogchap 
V8 `v8.9` now in stable channel.
https://omahaproxy.appspot.com/ shows the version is `8.9.255.20`

It supports top level `await`. https://v8.dev/blog/v8-release-89#top-level-await

Could you please update the V8 binary file version?

rogchap:
@kuoruan Top level await only works at the top level of a module not at the top top level; which is a bit miss leading.

It's still behind a feature flag but I'll look to see if we can add it and upgrade v8 to the latest stable version.

I'll do this in a seperate PR

promise.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Resolve
+func (r *PromiseResolver) Resolve(val Valuer) bool {
+	r.ctx.register()
+	defer r.ctx.deregister()
+	return C.PromiseResolverResolve(r.ptr, val.value().ptr) != 0
+}
+
+// Reject
+func (r *PromiseResolver) Reject(err *Value) bool {
```
kuoruan, line 69: Why not change this to Valuer?
So that we can reject an object here

lebronyang1:
for prom.State() == v8go.Pending {
	continue
}

@rogchap Hi, I don't always wait for promise results, Are there more effective ways to resolve this problem? 
Wether I can allows the isolate execute other context, when wait promise results? In other words, How do I use the isolate in multiple threads? Especially promise, I think it's so slowly.

PR #78, opened 2021-02-16
Release value memory on context close

rogchap:
See https://gophers.slack.com/archives/C01FW4CR5D2/p1613070695024600 for full details.

TLDR; When returning a value to Go from a V8 Context the Persistent Value storage was not being Reset correctly, and also stopped the V8 Context from being GC'd. Added memory issues where also caused by Go's GC not firing fast enough as it does not understand the memory allocated to the objects held by pointers in the C world.

This fix, tracks Values returned from the context, and when the Context is closed (either manually or by the Go GC) it will also reset all those values correctly.

### DIFF
--- a/v8go.cc
+++ b/v8go.cc
@@ -11,6 +11,7 @@
 #include <iostream>
 #include <sstream>
 #include <string>
+#include <vector>
 
 #include "libplatform/libplatform.h"
 #include "v8.h"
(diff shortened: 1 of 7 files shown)


tmc (approved):
This looks reasonable to me but I'm not pretending to be an expert here.

v8go.cc
```diff
@@ -93,6 +95,28 @@ RtnError ExceptionError(TryCatch& try_catch, Isolate* iso, Local<Context> ctx) {
...
+  // unordered_map, where we could do O(1) lookups for the value, but then know
+  // if the object has been finalized or not by being in the map or not. This
+  // would require some ref id for the value rather than passing the ptr between
+  // Go <--> C, which would be a significant change, as there are places where
+  // we get the context from the value, but if we then need the context to get
+  // the value, we would be in a circular bind.
+  ValuePtr val_ptr = static_cast<ValuePtr>(val);
+  ctx->vals.push_back(val_ptr);
+
+  return static_cast<ValuePtr>(val);
```
zwang, line 117: Does this create a new ValuePtr? Is it the same as `val_ptr` above or actually meant to be different?

rogchap: It does not create a new `ValuePtr`; `ValuePtr` is an alias to `void *` aka a pointer. 
So this is just doing the same thing as L114. 
Thanks for picking this up; although not effecting the code, it is a typo and I should just return `val_ptr` for simplicity.

Issue #81, opened 2021-02-26
[bug] segmentation violation panic after garbage collection


inn0kenty:
Hello!

I investigate random segmentation violation panics. This happens when GC execute value finalizer on disposed isolate. 

Maybe related to #39.

Code to reproduce.

```go
package main

import (
	"fmt"
	"runtime"
	"time"

	"rogchap.com/v8go"
)

func main() {
	iso, _ := v8go.NewIsolate()
	v8ctx, _ := v8go.NewContext(iso)
	val, _ := v8go.NewValue(iso, "some string")

	fmt.Println(val.String())

	v8ctx.Close()
	iso.Dispose()

	runtime.GC() // force garbage collection

	time.Sleep(time.Second)
}
``` 

Stack trace

```
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4232ee8]

runtime stack:
runtime.throw(0x4c066c4, 0x2a)
        /usr/local/Cellar/go/1.16/libexec/src/runtime/panic.go:1117 +0x72
runtime.sigpanic()
        /usr/local/Cellar/go/1.16/libexec/src/runtime/signal_unix.go:718 +0x2ef

goroutine 5 [syscall]:
runtime.cgocall(0x40adde0, 0xc00004ee90, 0xf)
        /usr/local/Cellar/go/1.16/libexec/src/runtime/cgocall.go:154 +0x5b fp=0xc00004ee60 sp=0xc00004ee28 pc=0x400657b
rogchap.com/v8go._Cfunc_ValueFree(0xcb185e0)
        _cgo_gotypes.go:618 +0x3c fp=0xc00004ee90 sp=0xc00004ee60 pc=0x40aa91c
rogchap.com/v8go.(*Value).finalizer.func1(0xc0000122e0)
        /Users/user/projects/project/vendor/rogchap.com/v8go/value.go:530 +0x55 fp=0xc00004eec8 sp=0xc00004ee90 pc=0x40ad155
rogchap.com/v8go.(*Value).finalizer(0xc0000122e0)
        /Users/user/projects/project/vendor/rogchap.com/v8go/value.go:530 +0xa5 fp=0xc00004ef38 sp=0xc00004eec8 pc=0x40ac4a5
runtime.call16(0x0, 0x4c07d70, 0xc00001a070, 0x1000000010)
[...]

PR #82, opened 2021-03-02
Fix value free panic on disposed isolate

inn0kenty:
See #81.

Possible solution with atomic "isolate disposed" flag.

### DIFF
--- a/context.go
+++ b/context.go
@@ -115,7 +115,7 @@ func (c *Context) RunScript(source string, origin string) (*Value, error) {
 // global proxy object.
 func (c *Context) Global() *Object {
 	valPtr := C.ContextGlobal(c.ptr)
-	v := &Value{valPtr, c}
+	v := &Value{valPtr, c, nil}
 	return &Object{v}
 }
 
@@ -175,7 +175,7 @@ func getValue(ctx *Context, rtn C.RtnValue) *Value {
 	if rtn.value == nil {
 		return nil
 	}
-	return &Value{rtn.value, ctx}
+	return &Value{rtn.value, ctx, nil}
 }
 
 func getError(rtn C.RtnValue) error {
--- a/isolate.go
+++ b/isolate.go
@@ -10,6 +10,7 @@ import "C"
 import (
 	"runtime"
 	"sync"
+	"sync/atomic"
 )
 
 var v8once sync.Once
@@ -20,9 +21,10 @@ var v8once sync.Once
 type Isolate struct {
 	ptr C.IsolatePtr
 
-	cbMutex sync.RWMutex
-	cbSeq   int
-	cbs     map[int]FunctionCallback
+	cbMutex  sync.RWMutex
+	cbSeq    int
+	cbs      map[int]FunctionCallback
+	disposed *int32
 }
 
 // HeapStatistics represents V8 isolate heap statistics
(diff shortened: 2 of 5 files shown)


rogchap:
I don't think the issue is with the Isolate being disposed as this still panics without disposing the Isolate before the GC.
I've tracked it down to the way the value is created against the Isolate, but most values are created on the context.
We use an internal Context for BigInt values, so I will see if tracking new values against this internal context is better.

This is my current test cases that panics:

```go
func TestIsolateGarbageCollection(t *testing.T) {
	t.Parallel()

	iso, _ := v8go.NewIsolate()
	val, _ := v8go.NewValue(iso, "some string")
	fmt.Println(val.String())

	// iso.Dispose()

	runtime.GC() // force garbage collection

	time.Sleep(time.Second)
}
```

Thanks for taking a look into this.

rogchap:
I need to fix up the CI for FOSSA too 😱

inn0kenty:
Thank you for your work!

I suppose your test case also panics because GC execute isolate finalizer before value finalizer.

Try comment line https://github.com/rogchap/v8go/blob/9848a00f5cb55a1da3ec045dec61d353c4f98620/isolate.go#L103

rogchap:
> I suppose your test case also panics because GC execute isolate finalizer before value finalizer.

Good point. 👍 

I'm thinking that using finalizes are just a bad idea, and too hard to coordinate between the C and Go worlds. 

When we create a new value, it's not attached to a context, but in order to return the contents of the value we need a context; so currently we use a single "internal" context for this.

With a regular user created Context, all values that are created on this context are attached to said context, so when we do `ctx.Close()` we release all the values that are tracked with that context.

I think for the new values that are created against just the Isolate, we should attach these values to the internal context vector. This way when we call `iso.Dispose()` we will free all the values created with this Isolate and then close the internal context, all be before disposing of the isolate.

This would eliminate the need for any finalizers on values, or contexts (you would HAVE To use `ctx.Close()` to avoid leaks), simplifying the Go code.

inn0kenty:
I think removing finalizers is a good point.

Also as I see Value already have Context pointer. So, why NewValue can't take context instead of isolate as a first argument? In that case will be possible track values through provided context instead of internal.

https://github.com/rogchap/v8go/blob/794781e582f89ec88eed4f089cdc32c531e2940f/value.go#L22
https://github.com/rogchap/v8go/blob/794781e582f89ec88eed4f089cdc32c531e2940f/value.go#L44

rogchap:
> Also as I see Value already have Context pointer. So, why NewValue can't take context instead of isolate as a first argument? In that case will be possible track values through provided context instead of internal.

Yea, it's slightly complicated with the V8 API: to "resolve" any Value (ie get the string or integer value) you require a Context, but to just create a Value you only need the Isolate. The reason for this is that you need to be able to set properties of the Object/Function Templates that can be use to create a context.
It would be counter intuitive to require a Context to create a Context.

FYI: I've created #86 which should fix your issue described here. I tested the finalizer for Templates and it seems fine/not effected, by disposing the Isolate.

Are you able to test your code against this fix?

inn0kenty:
Thank you!

This fix works fine with my code.

rogchap:
Closing as this was fixed in #86

PR #100, opened 2021-04-03
Support for invoking a function value

### LINKED ISSUE #95: Invoke a v8go.Value ?
Hi,
If I have a v8go.Value where `IsFunction()` returns `true`, is there any way to invoke that value? 
Thank you!
Rob

robfig:
Fixes #95

### DIFF
--- a/function.go
+++ b/function.go
@@ -16,7 +16,7 @@ type Function struct {
 }
 
 // Call this JavaScript function with the given arguments.
-func (fn *Function) Call(args []Valuer) (*Value, error) {
+func (fn *Function) Call(args ...Valuer) (*Value, error) {
 	var argptr *C.ValuePtr
 	if len(args) > 0 {
 		var cArgs = make([]C.ValuePtr, len(args))
--- a/function.go
+++ b/function.go
@@ -1,0 +1,31 @@
+// Copyright 2021 Roger Chapman and the v8go contributors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package v8go
+
+// #include "v8go.h"
+import "C"
+import (
+	"unsafe"
+)
+
+// Function is a JavaScript function.
+type Function struct {
+	*Value
+}
+
+// Call this JavaScript function with the given arguments.
+func (fn *Function) Call(args []*Value) (*Value, error) {
+	var rtn C.RtnValue
+	if len(args) == 0 {
+		rtn = C.FunctionCall(fn.ptr, C.int(0), (*C.ValuePtr)(nil))
+	} else {
+		cArgs := make([]C.ValuePtr, len(args))
+		for i, arg := range args {
+			cArgs[i] = arg.ptr
+		}
+		rtn = C.FunctionCall(fn.ptr, C.int(len(args)), (*C.ValuePtr)(unsafe.Pointer(&cArgs[0])))
+	}
+	return getValue(fn.ctx, rtn), getError(rtn)
+}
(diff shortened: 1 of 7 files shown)


robfig:
This is a work in progress -- still need to add more tests -- but let me know if I'm not on the right track.

Thanks!

robfig:
OK, this should have 100% test coverage. The current failure is about a missing FOSSA key(?)

tmc (approved):
This all looks reasonable to me. @rogchap ?

function.go
```diff
@@ -0,0 +1,31 @@
...
+	"unsafe"
+)
+
+// Function is a JavaScript function.
+type Function struct {
+	*Value
+}
+
+// Call this JavaScript function with the given arguments.
+func (fn *Function) Call(args []*Value) (*Value, error) {
```
rogchap, line 19: Can we change the `args` to be the `Valuer` interface? Then you can use any Value type. ie Object etc

function.go
```diff
@@ -0,0 +1,31 @@
...
+	var rtn C.RtnValue
+	if len(args) == 0 {
+		rtn = C.FunctionCall(fn.ptr, C.int(0), (*C.ValuePtr)(nil))
+	} else {
```
rogchap, line 23: Is this else statement needed? The range would still work with a empty array?
If we need to pass the nil pointer can we return in the above `if` statement to avoid the extra branching caused by the `else`

rogchap (changes requested):
This looks great. Thank you for putting this PR together, just a few minor comments/changes

function.go
tmc: Good catch!

function.go
robfig: That was something I found by writing the test -- to get the `*C.ValuePtr`, I have to take the value of the first element, and that is an illegal index on an empty slice. 

Updated to avoid calling FunctionCall in two different branches.

function.go
robfig: Updated

robfig:
Thanks for the review. Comments addressed, and tests pass locally.

function_test.go
```diff
@@ -0,0 +1,63 @@
...
+	failIf(t, err)
+	addValue, err := ctx.Global().Get("add")
+	failIf(t, err)
+	iso, _ := ctx.Isolate()
+
+	arg1, err := v8go.NewValue(iso, int32(1))
+	failIf(t, err)
+
+	fn, _ := addValue.AsFunction()
+	resultValue, err := fn.Call([]v8go.Valuer{arg1, arg1})
```
rogchap, line 28: This is looking great @robfig.
Sorry to be a pain, but looking at this I'm wondering if a variadic function would be more idiomatic? Then you could do:
```
fn.Call(arg1, arg2, etc)
```

rogchap (changes requested):
I should have mentioned this in the first review, sorry for the back and forth. Please could you also update the `CHANGELOG.md`? Don't forget to add your attribution 💪

function_test.go
robfig: No problem at all -- agreed that it looks cleaner. Updated

robfig:
Added & updated!

rogchap:
Perfect @robfig 
Thanks again. I'll fix up the FOSSA error (GitHub secrets don't work in forks)

rogchap (approved):

robfig:
Sounds good -- not sure if you expect me to merge, but I don't have access to do that despite your approval. I'll leave it to you.

rogchap:
I'll merge shortly.

PR #101, opened 2021-04-03
Context.RunModule: execute ES Module source

### LINKED ISSUE #99: Support for ES Modules
It looks like v8go does not support ES Modules. For example, I translated the Context example to use modules:
```
package main

import (
	"fmt"

	"rogchap.com/v8go"
)

func main() {
	ctx, _ := v8go.NewContext()
	val, err := ctx.RunScript("export function add(a, b) { return a + b };", "math.mjs")
	fmt.Println(val, err)
	val, err = ctx.RunScript("import { add } from 'math.mjs';\nconst result = add(3, 4)", "main.js")
	fmt.Println(val, err)
	val, err = ctx.RunScript("result", "value.js")
	fmt.Println(val, err)
	fmt.Println(val)
}
```

and it resulted in these errors:
```
<nil> SyntaxError: Unexpected token 'export'
<nil> SyntaxError: Cannot use import statement outside a module
...
```

I do not typically develop JS or use Node.js, so I might be doing something dumb too. 

Do you have any particular ideas around how this might work? I'd be happy to submit a pull request if so.

Thanks!
Rob

robfig:
The result is a Value of type ModuleNamespace, from which the module's
exports can be accessed. Imports are not supported.

Fixes #99

### DIFF
--- a/v8go.cc
+++ b/v8go.cc
@@ -414,6 +414,86 @@ RtnValue RunScript(ContextPtr ctx_ptr, const char* source, const char* origin) {
   return rtn;
 }
 
+static MaybeLocal<Module> ResolveCallback(Local<Context> context,
+                                          Local<String> specifier,
+                                          Local<Module> referrer) {
+  Isolate* isolate = context->GetIsolate();
+  isolate->ThrowException(String::NewFromUtf8(isolate, "import not supported").ToLocalChecked());
+  return MaybeLocal<Module>();
+}
+
+RtnValue RunModule(ContextPtr ctx_ptr, const char* source, const char* origin) {
+  LOCAL_CONTEXT(ctx_ptr);
+
+  Local<String> src =
+      String::NewFromUtf8(iso, source, NewStringType::kNormal).ToLocalChecked();
+  Local<String> ogn =
+      String::NewFromUtf8(iso, origin, NewStringType::kNormal).ToLocalChecked();
+  RtnValue rtn = {nullptr, nullptr};
+
+  ScriptOrigin script_origin(ogn, 0, 0, false, -1, Local<Value>(), false, false, true);
+  ScriptCompiler::Source csrc(src, script_origin);
+  MaybeLocal<Module> mmodule = ScriptCompiler::CompileModule(iso, &csrc);
+  if (mmodule.IsEmpty()) {
+    rtn.error = ExceptionError(try_catch, iso, local_ctx);
+    return rtn;
+  }
+
+  Local<Module> root_module = mmodule.ToLocalChecked();
+  if (!root_module->InstantiateModule(local_ctx, &ResolveCallback).FromMaybe(false)) {
+    rtn.error = ExceptionError(try_catch, iso, local_ctx);
+    return rtn;
+  }
+
+  Local<Value> result;
+  if (!root_module->Evaluate(local_ctx).ToLocal(&result)) {
+    rtn.error = ExceptionError(try_catch, iso, local_ctx);
+    return rtn;
+  }
+
+  // result is a Promise if the module has imports that need to be resolved
+  // OR has a top-level await. imports are not supported yet, but awaits are.
+  //
+  // If await was not used, the result is undefined, and the module's status
+  // should already be kEvaluated.
+  if (result->IsPromise()) {
+    // We have to wait. Just busy-loop and poll for completion.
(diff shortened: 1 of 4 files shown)


robfig:
This builds upon the other FunctionCall PR. Do you know if there's a convention around stacks of PRs? This diff now contains both changes, which isn't desirable.

robfig:
Oh, I should say that this results in a cgo warning that I couldn't figure out how to get rid of. Any tips?

```
v8go.cc:412:20: warning: 'ResolveCallback' has C-linkage specified, but returns incomplete type 'MaybeLocal<v8::Module>' which could be incompatible with C [-Wreturn-type-c-linkage]
```

robfig:
In testing this with top-level await, I found that I had to handle modules which were not synchronously evaluated. I added handling for that along with some more tests.

cleiner:
> 
> 
> Oh, I should say that this results in a cgo warning that I couldn't figure out how to get rid of. Any tips?
> 
> ```
> v8go.cc:412:20: warning: 'ResolveCallback' has C-linkage specified, but returns incomplete type 'MaybeLocal<v8::Module>' which could be incompatible with C [-Wreturn-type-c-linkage]
> ```

The compiler warning is about having a non-static function with templated (C++) types in an extern "C" block -- as written it looks like it's supposed to be part of the API exposed to cgo, although it would (probably) not be callable.
Declaring the function as `static MaybeLocal<Module> ResolveCallback(...)` limits its scope to v8go.cc and should fix this.

robfig:
Thanks @cleiner , fixed! Confirmed that it got rid of the warning and all tests continue to pass.

rogchap:
Hey @robfig are you able to rebase with master and resolve the conflicts?

robfig:
Rebased and merged

robfig:
I'm starting to second guess this interface. 

Although Go idiomatically exposes blocking, synchronous interfaces and leaves it to the caller to make things async when there's a need, in this case the underlying interface is async. When we support module loading, RunModule might block for a long time, and even without imports the module might use a top-level await which does something time consuming. 

Since the blocking happens in C code, it will consume a thread, unlike blocking in Go code. But on the flip side, exposing the async nature of the interface would increase the interface's complexity, and it would probably lead to more back-and-forths between Go and C which I suppose is relatively costly (although maybe at these scales). 

Any thoughts?

robfig:
The underlying result of "running a module" is either immediate completion or a Promise, if there's a top-level await or imports to resolve. We could do something similar in the Go interface and return a Value which is either a Promise or a ModuleNamespace.

However, it looks like the Promise we expose is not able to blocked upon.. without that it seems like it might be sort of onerous to get a Promise back from RunModule, because the caller would have to poll it for completion.

v8go.cc
```diff
@@ -414,6 +414,86 @@ RtnValue RunScript(ContextPtr ctx_ptr, const char* source, const char* origin) {
...
+  // OR has a top-level await. imports are not supported yet, but awaits are.
+  //
+  // If await was not used, the result is undefined, and the module's status
+  // should already be kEvaluated.
+  if (result->IsPromise()) {
+    // We have to wait. Just busy-loop and poll for completion.
+    Local<Promise> result_promise(Local<Promise>::Cast(result));
+    while (result_promise->State() == Promise::kPending) {
+      iso->PerformMicrotaskCheckpoint();
+    }
```
dylanahsmith, line 464: The top-level await could be blocked on IO done in another go-routine, in which case this spin loop will just waste the thread's CPU time.  The other problem with this loop is that broken promises (those neither resolved or rejected) would result in an infinite loop.

I think the function should just return the promise, which would be consistent with how RunScript doesn't try to wait for promises either.

robfig: I agree, the result of RunModule should be a Promise. I did some work towards enabling that in https://github.com/rogchap/v8go/pull/119 and https://github.com/rogchap/v8go/pull/118 , but since then my project has adopted an alternative approach to what I was using v8 for, so I'm probably not going to continue this PR. Closing.

RobertWHurst:
I'm kind of supersized that no one else picked this up after the fact. Doesn't seem like there is much left to do. @dylanahsmith what's your take here, it's been some time since this was worked on. What would need to be done to complete this functionality apart from returning a promise value rather than trying to handle the result in c?

PR #102, opened 2021-04-03
Build for static V8 library on Windows

cleiner:
This PR adds the bits required to provide a prebuilt static V8 library on Windows, offering basically the same level of end user convenience as on Linux/Mac (i.e. not having to deal with a dozen DLLs in a distribution).

Existing Windows builds using v8go should not break, the change merely renders the need to package additional files with the executable unnecessary.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         go-version: [1.12.17, 1.15.7]
-        platform: [ubuntu-latest, macos-latest]
+        platform: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.platform }}
     
     steps:
@@ -23,6 +23,10 @@ jobs:
         go-version: ${{ matrix.go-version }}
     - name: Checkout
       uses: actions/checkout@v2
+    - name: Add MINGW to PATH
+      if: matrix.platform == 'windows-latest'
+      run: echo "C:\msys64\mingw64\bin" >> $GITHUB_PATH
+      shell: bash
     - name: Go Test
       run: go test -v -coverprofile c.out ./...
     - name: Upload coverage to Codecov
--- a/.github/workflows/v8build.yml
+++ b/.github/workflows/v8build.yml
@@ -7,7 +7,7 @@ jobs:
         name: Build V8 for ${{ matrix.platform }}
         strategy:
             matrix:
-                platform: [ubuntu-latest, macos-latest]
+                platform: [ubuntu-latest, macos-latest, windows-latest]
         runs-on: ${{ matrix.platform }}
         steps:
             - name: Checkout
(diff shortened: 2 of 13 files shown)


neptoess:
@cleiner ,
This is pretty cool, but I'm wondering if it makes more sense to try and get this merged into MSYS2, so _everyone_ using their package has access to a static `libv8.a`, not just people using v8go. This also removes the need for this repo to track MSYS2's patches, and makes bundling the static library a simple copy command.

cleiner:
@neptoess,
I had similar thoughts along the way, but more in terms of asking the MSYS2 maintainers if they had tried submitting their patches to Chromium/V8, as it did support MINGW in the past, before the build system was migrated to GN (though I'm a little concerned this will turn out to be kind of a long term endeavor, if at all feasible).

With regard to merging this into MSYS2, I think there's nothing to merge here (except for the tiny `0002` patch) -- it would be more like asking for an additional package built to v8go's needs:
- Changing the existing package would be breaking for the people currently depending on it
- ICU is included in the MSYS2 build, but not in v8go; adding it would make the binary too big to bundle
- v8_untrusted_code_mitigations is enabled for MSYS2, but not for v8go (don't know about the consequences of changing this)
- I think the component build was chosen deliberately, so people can create their own snapshot_blobs

As I understand it, the need to track the MSYS2 package already exists (and would continue to exist even with a provided libv8.a), because if v8go uses a newer V8 version than MSYS2, Windows builds may break as the V8 API is not guaranteed to stay compatible, is it?
If this is the case, then versions shouldn't diverge and copying the patches will be one additional step in the "Upgrading the V8 binaries" section of the readme. There'd be no need to "patch the patches" as I did here.

That said, I get that it's a potential maintenance burden in its current form and I'd understand if the project would rather not take it on.

neptoess:
Good info @cleiner. I agree this is the path to take. I'll leave it up to @rogchap and the other maintainers to determine whether they want to bundle a static lib for Windows. I personally think it's a good idea though.

rogchap:
@cleiner WOW This looks amazing 🤩 
Will make so many Windows users happy.
Will take a closer look this afternoon.

rogchap:
@zwang Do you have any time to look at this one, as you are more familiar with the Window ecosystem than I?

I want to make sure the path to upgrading V8 to latter versions will be possible/easy/documented, which is my main concern.

rogchap:
@cleiner can you possibly update with latest master? For some reason GH is returning a 500 when I try and use the GUI to try and update.
I've fixed the FOSSA issue, so the tests should now run successfully.

zwang:
> @zwang Do you have any time to look at this one, as you are more familiar with the Window ecosystem than I?
> 
> I want to make sure the path to upgrading V8 to latter versions will be possible/easy/documented, which is my main concern.

The overall changes looks good to me. Although I could not verify the correctness of the patches, I assume the patches are good since they are from mingw. 

Regarding the path to upgrading V8 to latter versions, it seems to be possible and is well documented as I see.

zwang (approved):

cleiner:
> 
> 
> @cleiner can you possibly update with latest master? For some reason GH is returning a 500 when I try and use the GUI to try and update.
> I've fixed the FOSSA issue, so the tests should now run successfully.

@rogchap Rebased on current master.

PR #104, opened 2021-04-09
Fix FOSSA scan on Windows

cleiner:
@rogchap Missed that the FOSSA step would break the Windows build in #102, sorry.

If the scan doesn't have to run on Windows and can be skipped, feel free to close this PR (and just copy the skip condition from line 43).
Otherwise the separate step using Powershell is required, because the FOSSA install script doesn't work in Git bash on Windows.


zwang (approved):
Thanks 👍

Issue #53, opened 2020-12-21
libv8.a bundled with v0.3.0 doesn't work with golang Docker image


neptoess:
Testing with a very basic `main.go`
```go
package main

import "rogchap.com/v8go"

func main() {
	if _, err := v8go.NewIsolate(); err != nil {
		panic(err)
	}
}
```

And the following Dockerfile
```
FROM golang:1.15

WORKDIR /go/src/v8
COPY . .
RUN go run main.go
```

I get this from `docker build .`
```
Sending build context to Docker daemon  75.76MB
Step 1/4 : FROM golang:1.15
 ---> 05c8f6d2538a
Step 2/4 : WORKDIR /go/src/v8
 ---> Using cache
 ---> 70761780e417
Step 3/4 : COPY . .
 ---> 96c01a616e96
Step 4/4 : RUN go run main.go
 ---> Running in f434a5b958aa
# rogchap.com/v8go
/usr/bin/ld: vendor/rogchap.com/v8go/deps/linux-x86_64/libv8.a(isolate.o): in function `std::_Sp_counted_ptr_inplace<v8::internal::Counters, std::allocator<v8::internal::Counters>, (__gnu_cxx::_Lock_policy)2>::_M_get_deleter(std::type_info const&)':
isolate.cc:(.text._ZNSt23_Sp_counted_ptr_inplaceIN2v88internal8CountersESaIS2_ELN9__gnu_cxx12_Lock_policyE2EE14_M_get_deleterERKSt9type_info[_ZNSt23_Sp_counted_ptr_inplaceIN2v88internal8CountersESaIS2_ELN9__gnu_cxx12_Lock_policyE2EE14_M_get_deleterERKSt9type_info]+0x16): undefined reference to `std::_Sp_make_shared_tag::_S_eq(std::type_info const&)'
/usr/bin/ld: vendor/rogchap.com/v8go/deps/linux-x86_64/libv8.a(isolate.o): in function `std::_Sp_counted_ptr_inplace<v8::internal::metrics::Recorder, std::allocator<v8::internal::metrics::Recorder>, (__gnu_cxx::_Lock_policy)2>::_M_get_deleter(std::type_info const&)':
[...]

PR #106, opened 2021-04-14
test: run code test on more platforms

kuoruan:
platform versions are from: https://github.com/actions/virtual-environments

### DIFF
--- a/.github/workflows/gcc_test.yml
+++ b/.github/workflows/gcc_test.yml
@@ -1,0 +1,30 @@
+name: GCC Test
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  gcc:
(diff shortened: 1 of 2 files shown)


kuoruan:
All the GCC versions check have passed 😂, any thing I did wrong?

cleiner:
@kuoruan, my earlier comment in #53 was based on verifying the following:
- libv8.a from master as is does _not_ work with the golang:1.15 image
- libv8.a from rebuilding master on ubuntu-18.04 works
- libv8.a from v0.5.1 works
- libv8.a from rebuildung v0.5.1 on ubuntu-latest does _not_ work
- libv8.a from master as is works with the golang:1.15 image rebuilt `FROM buildpack-deps:focal-scm` (i.e. 20.04, same as ubuntu-latest)

So while it may not necessarily be GCC itself, there's some environment-dependent version/ABI incompatibility (libstdc++?).

PR #128, opened 2021-05-05
Correct example in readme

genevieve:


rogchap (approved):
Thanks for spotting this.

PR #133, opened 2021-05-09
Assign concrete err val to new var in readme

genevieve:
Forgot this in my previous pr


rogchap (approved):
Thanks; didn't even pick this up 😬

PR #134, opened 2021-05-14
Add Function.NewInstance to expose the corresponding V8 function

dylanahsmith:
## Problem

I was trying to do something like `new Error(message)` from Go code, but didn't see a way of using the javascript `new` operator.

## Solution

It looks like v8::Function::NewInstance is the function we need for doing this, so I added a corresponding function to the v8go API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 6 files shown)


rogchap (approved):

rogchap:
Hey @dylanahsmith 
Thanks for this PR. Code looks good; would be good if we could raise the test coverage.
Could you also add an entry to the CHANGELOG.md?

dylanahsmith:
I've added an extra test for the error case and a changelog entry.  Looks like CI is awaiting approval to run

PR #136, opened 2021-05-24
Build v8 with i18n support

dylanahsmith:
## Problem

I was getting an `ReferenceError: Intl is not defined` error from trying to use the Intl JS namespace which is a builtin feature of v8 (https://v8.dev/blog/intl).  I found this wasn't working because v8go was building with `v8_enable_i18n_support=false`.

## Solution

Removing `v8_enable_i18n_support=false` from the build configuration and rebuilding solved the problem for me.

As instructed in the README.md, I didn't add any built binaries to the commit I pushed.  I don't have access to run the "V8 Build" Github Action, so that still needs to be done to update the pre-built binaries in order to actually fix this issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Support for calling constructors functions with NewInstance on Function
+- Build v8 with i18n support
 
 ## [v0.6.0] - 2021-05-11
 
--- a/deps/build.py
+++ b/deps/build.py
@@ -51,7 +51,6 @@ v8_use_external_startup_data=false
 treat_warnings_as_errors=false
 v8_embedder_string="-v8go"
 v8_enable_gdbjit=false
-v8_enable_i18n_support=false
 v8_enable_test_features=false
 v8_untrusted_code_mitigations=false
 exclude_unwind_tables=true


rogchap:
Thanks @dylanahsmith 
I’ll trigger the build of V8 for the updated binaries.
I originally excluded this to reduce the V8 file size, so will be keen to see what this adds.

rogchap (approved):

nsf:
May I also add some info in case somebody decides to pursue this feature. On my system V8 was crashing when attempting to use Intl, even when built with i18n. Had to add this line:

```cpp
v8::V8::InitializeICU("/path/to/icudtl.dat");
```

rogchap:
@nsf The master branch does not have the binaries built for i18n because it failed to build on Windows.
You would need to build the v8 lib locally yourself.

nsf:
@rogchap Yes, I'm aware of that. I was just mentioning that when you build v8 with i18n enabled, in addition to that you have to point it at icudtl.dat file, which contains all the info about how to format things in different locales. And you do so by calling `V8::InitializeICU` with the right path to icudtl.dat file. By default v8 build is not shipped with it. There are some ways to embed it, but it was not trivial, so I skipped figuring it out.

Well, if you google it, it's easy to find struggles in other related projects (like node or deno), e.g.: https://github.com/denoland/deno/issues/1968

Just saying it's slightly less trivial than just enabling it via build option. However, these days libs like https://moment.github.io/luxon/ appear (which rely on Intl) and having Intl support becomes a must.

nsf:
Also, I just noticed that when you build it without clang, the ICU data file ends up being embbeded to `libv8.a`. Maybe that was the problem for those of us who tried building it ourselves to begin with. To build it without clang:

```
./build.py --no-clang
```
(README.md didn't mention it)

I can indeed confirm that when you build it with clang there is no such file in there, but with gcc it's there:
```sh
# clang build:
> ls -l libv8.a
-rw-r--r-- 1 nsf users 54662470 Aug 17 11:06 libv8.a

> ar -tv libv8.a | grep icudtl
Exception: grep exited with 1

# gcc build:
> ls -l libv8.a
-rw-r--r-- 1 nsf users 73969530 Aug 24 09:15 libv8.a

> ar -tv libv8.a | grep icudtl
rw-r--r-- 0/0 10414384 Jan  1 05:00 1970 icudtl_dat.o
```

PR #138, opened 2021-06-05
Revert "Build v8 with i18n support"

rogchap:
Reverts rogchap/v8go#136


rogchap:
@dylanahsmith 
I tried to create a build without this flag, but the windows build consistently failed.
I'm re-running the build against this PR to confirm that this is the cause.

goenning:
It seems like some recent windwos builds are passing so maybe this is no the issue? I'm hoping this won't be reverted as I'm interested on using Intl as well!

PR #142, opened 2021-06-08
Add access to "this" from function callback

### LINKED ISSUE #122: How to access `this` within a function template?
You can use a FunctionTemplate to create a constructor, having it return an instance from an ObjectTemplate.

You can set functions *on* an ObjectTemplate so the object has methods.

But, when someone calls a method (impl as a function template) on an object (from an object template), there's no way that I can see to get access to the object itself from within the method implementation. Are there any examples of something like this actually working?

I started a [thread on slack](https://gophers.slack.com/archives/C01FW4CR5D2/p1618946801073100) but haven't heard anything.

Code example:

```go
package main

import (
	"fmt"

	"rogchap.com/v8go"
)

func main() {
	iso, _ := v8go.NewIsolate()
	g, _ := v8go.NewObjectTemplate(iso)

	ot, err := v8go.NewObjectTemplate(iso)
	if err != nil {
		panic(err)
	}

	fooFn, err := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *v8go.Value {
		// How do I get the instance created in the constructor below?
		fmt.Printf("MyObject.foo(): %q\n", info.Args())
		v, _ := v8go.NewValue(iso, "what is 'this'")
		return v
	})
	if err != nil {
		panic(err)
	}

	if err := ot.Set("foo", fooFn); err != nil {
		panic(err)
	}

	constructor, _ := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *v8go.Value {
		inst, _ := ot.NewInstance(info.Context())
		inst.Set("prop", info.Args()[0])
		return inst.Value
	})

	if err := g.Set("MyObject", constructor); err != nil {
		panic(err)
	}

	ctx, _ := v8go.NewContext(iso, g)

	rv, err := ctx.RunScript(`let x = new MyObject('bar'); x.foo()`, "script.js")
	fmt.Printf("rv: %q, err: %q\n", rv, err)
}
```

rogchap:
fixes #122

### DIFF
--- a/function_template_test.go
+++ b/function_template_test.go
@@ -79,6 +79,33 @@ func TestFunctionTemplateGetFunction(t *testing.T) {
 	}
 }
 
+func TestFunctionCallbackInfoThis(t *testing.T) {
+	t.Parallel()
+
+	iso, _ := v8go.NewIsolate()
+
+	foo := v8go.NewObjectTemplate(iso)
+	foo.Set("name", "foobar")
+
+	var this *v8go.Object
+	barfn := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *v8go.Value {
+		this = info.This()
+		return nil
+	})
+	foo.Set("bar", barfn)
+
+	global := v8go.NewObjectTemplate(iso)
+	global.Set("foo", foo)
+
+	ctx, _ := v8go.NewContext(iso, global)
+	ctx.RunScript("foo.bar()", "")
+
+	v, _ := this.Get("name")
+	if v.String() != "foobar" {
+		t.Errorf("expected this.name to be foobar, but got %q", v)
+	}
+}
+
 func ExampleFunctionTemplate() {
 	iso, _ := v8go.NewIsolate()
 	global := v8go.NewObjectTemplate(iso)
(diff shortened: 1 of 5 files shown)


function_template_test.go
```diff
@@ -64,6 +66,36 @@ func TestFunctionTemplateGetFunction(t *testing.T) {
...
+
+	ctx, _ := v8go.NewContext(iso, global)
+	defer ctx.Close()
+
+	ctx.RunScript("foo.bar()", "")
+
+	v, _ := this.Get("name")
+	if v.String() != "foobar" {
+		t.Errorf("expected this.name to be foobar, but got %q", v)
+	}
```
dylanahsmith, line 96: Looks like this test would be simpler if we exposed [v8::Value::SameValue](https://v8.github.io/api/head/classv8_1_1Value.html#a7c90dc5947b8b99b9089a020fff6d86d) so we could test this with `if foo.SameValue(this)` without the need for "name" in this test.

rogchap:
Not sure why Windows is having a hard time with this code.

dylanahsmith:
> Not sure why Windows is having a hard time with this code.

Specifically, it seems to be having troubles with passing the extra argument to goFunctionCallback.  Specifically, I was able to reproduce this by just passing a NULL placeholder pointer instead of the rest of the code for this feature.  I'm not really sure why that is a problem, but a possible workaround seems to be passing the extra argument in the args array.  I've pushed a commit that does that and CI is now ✅ 

I'm still not sure what the underlying problem is, but I added a comment to document the problem to help capture the "why" of this workaround in the code.

dylanahsmith (approved):

PR #145, opened 2021-06-15
Implement set, get internal fields, & set internal field count

wisepythagoras:
I've implemented the `Object.SetInternalField`, `Object.GetInternalField`, and the `ObjectTemplate.SetInternalFieldCount` functions.

Is this a viable solution or am I missing something?

``` go
iso, _ := v8go.NewIsolate()
ctx, _ := v8go.NewContext(iso)

tmpl, _ := v8go.NewObjectTemplate(iso)
tmpl.SetInternalFieldCount(1)

obj, _ := tmpl.NewInstance(ctx)

obj.SetInternal(0, "baz")
v, err := obj.GetInternal(0)
```

### DIFF
--- a/object.go
+++ b/object.go
@@ -27,7 +27,7 @@ func (o *Object) Set(key string, val interface{}) error {
 	if len(key) == 0 {
 		return errors.New("v8go: You must provide a valid property key")
 	}
-	return set(o, key, 0, val)
+	return set(o, key, 0, val, false)
 }
 
 // Set will set a given index on the Object to a given value.
@@ -35,10 +35,15 @@ func (o *Object) Set(key string, val interface{}) error {
 // If the value passed is a Go supported primitive (string, int32, uint32, int64, uint64, float64, big.Int)
 // then a *Value will be created and set as the value property.
 func (o *Object) SetIdx(idx uint32, val interface{}) error {
-	return set(o, "", idx, val)
+	return set(o, "", idx, val, false)
 }
 
-func set(o *Object, key string, idx uint32, val interface{}) error {
+// SetInternal sets the value of an internal field.
+func (o *Object) SetInternal(idx uint32, val interface{}) error {
+	return set(o, "", idx, val, true)
+}
+
+func set(o *Object, key string, idx uint32, val interface{}, internal bool) error {
 	var value *Value
 	switch v := val.(type) {
 	case string, int32, uint32, int64, uint64, float64, bool, *big.Int:
@@ -58,7 +63,12 @@ func set(o *Object, key string, idx uint32, val interface{}) error {
 		return nil
 	}
 
-	C.ObjectSetIdx(o.ptr, C.uint32_t(idx), value.ptr)
+	if internal {
+		C.ObjectSetInternal(o.ptr, C.uint32_t(idx), value.ptr)
+	} else {
+		C.ObjectSetIdx(o.ptr, C.uint32_t(idx), value.ptr)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


v8go.cc
```diff
@@ -1010,6 +1031,25 @@ RtnValue ObjectGet(ValuePtr ptr, const char* key) {
   return rtn;
 }
 
+RtnValue ObjectGetInternal(ValuePtr ptr, uint32_t idx) {
```
wisepythagoras, line 1034: I can merge this with `ObjectGetIdx`, since they do the same thing, besides line 1038.

wisepythagoras:
Could I get a review on this as well, @dylanahsmith or @rogchap?

v8go.cc
```diff
@@ -1010,6 +1031,25 @@ RtnValue ObjectGet(ValuePtr ptr, const char* key) {
...
+  LOCAL_OBJECT(ptr);
+  RtnValue rtn = {nullptr, nullptr};
+
+  MaybeLocal<Value> result = obj->GetInternalField(idx);
```
dylanahsmith, line 1038: This returns a `Local<Value>`, so the `if (result.IsEmpty()) {` code path is dead code and we don't need to propagate an error for this call.

object.go
```diff
@@ -58,7 +63,16 @@ func set(o *Object, key string, idx uint32, val interface{}) error {
 		return nil
 	}
 
-	C.ObjectSetIdx(o.ptr, C.uint32_t(idx), value.ptr)
+	if internal {
+		inserted := C.ObjectSetInternal(o.ptr, C.uint32_t(idx), value.ptr)
+
+		if inserted == 0 {
+			return errors.New("v8go: index exceeded internal field count")
```
dylanahsmith, line 70: These fields are internal to host/go code, so getting and setting them should be more like getting and setting a go slice.  An index out of range error in Go is reported as a panic, so I think we should do the same here.  This is because the error isn't something that we expect the caller to handle, but more an indication of a bug in the caller. 
 Likely the caller would be using these fields more like struct field anyways.

If the library user happens to be using these internal fields as a bounded array, then we could expose `InternalFieldCount` so the assignment can be conditional (similar using the `len` function on a slice)

v8go.cc
```diff
@@ -989,6 +996,20 @@ void ObjectSetIdx(ValuePtr ptr, uint32_t idx, ValuePtr val_ptr) {
...
 
+int ObjectSetInternal(ValuePtr ptr, uint32_t idx, ValuePtr val_ptr) {
+  LOCAL_OBJECT(ptr);
+  m_value* prop_val = static_cast<m_value*>(val_ptr);
+
+  if (obj->InternalFieldCount() <= idx) {
+    return 0;
+  }
+
+  obj->SetAlignedPointerInInternalField(idx, nullptr);
```
dylanahsmith, line 1007: The documentation for SetAlignedPointerInInternalField says

> To retrieve such a field, GetAlignedPointerFromInternalField must be used, everything else leads to undefined behavior.

and there are no calls to GetAlignedPointerFromInternalField, so why is this needed?  If it is needed, then this line could use a comment clarifying its purpose and we could also use a test that would fail if someone tried to remove this line.

wisepythagoras: I should definitely add a comment. When I worked on this, I couldn't set an internal field unless I called this. And what I mean is that if I attempted to call `SetInternalField` without calling this, it would crash. I did some searching online and saw these two being used together, so I gave it a try and it worked.

That said, I'm open to correcting it, if it's not correct.

dylanahsmith: I'm not saying it isn't correct.  All I know is that removing this line doesn't cause a test to fail and that the purpose of the line wasn't clear when based on reading the v8 documentation for it.  If you could reproduce the crash in the test suite, it would be very helpful as a regression test, even if you don't understand why/how the line is needed to fix the crash.

wisepythagoras: Absolutely! I'm very curious, though. If it doesn't crash, this can definitely be removed, but I can't quite remember what exactly the issue was. I may have some notes, so I'll take a look.

Thanks for pointing this out.

wisepythagoras: I've removed this because I can't repro.

PR #147, opened 2021-06-24
Function -> Script Origin -> Source Map Url

genevieve:


function_test.go
```diff
@@ -33,6 +33,26 @@ func TestFunctionCall(t *testing.T) {
...
+
+	fn, _ := addValue.AsFunction()
+
+	resultVal, err := fn.SourceMapUrl()
+	failIf(t, err)
+
+	if resultVal.String() != "main.js.map" {
+		t.Errorf("expected main.js.map, got %v", resultVal.String())
+	}
+}
```
dylanahsmith, line 54: Please add a test for when there is no source map url

function.go
```diff
@@ -46,3 +46,11 @@ func (fn *Function) NewInstance(args ...Valuer) (*Object, error) {
 	fn.ctx.deregister()
 	return getObject(fn.ctx, rtn), getError(rtn)
 }
+
+// Return the script origin for a function
+func (fn *Function) SourceMapUrl() (*Value, error) {
+	fn.ctx.register()
+	rtn := C.FunctionSourceMapUrl(fn.ptr)
+	fn.ctx.deregister()
```
dylanahsmith, line 54: It doesn't look like we need to lookup the v8go.Context from the v8::Context in `FunctionSourceMapUrl`, so I don't think the register and deregister is necessary.  I think it is only necessary in the above functions because JS code is being called that could then call a Go callback.

```suggestion
	rtn := C.FunctionSourceMapUrl(fn.ptr)
```

v8go.cc
```diff
@@ -1207,6 +1207,19 @@ RtnValue FunctionNewInstance(ValuePtr ptr, int argc, ValuePtr args[]) {
   return rtn;
 }
 
+RtnValue FunctionSourceMapUrl(ValuePtr ptr) {
+  LOCAL_VALUE(ptr)
+  RtnValue rtn = {nullptr, nullptr};
+  Local<Function> fn = Local<Function>::Cast(value);
+  MaybeLocal<Value> result = fn->GetScriptOrigin().SourceMapUrl();
```
dylanahsmith, line 1214: It looks like [ScriptOrigin::SourceMapUrl](https://v8docs.nodesource.com/node-16.0/db/d84/classv8_1_1_script_origin.html#a1d2ec57af7279bd782132fcdff522bc6) returns a `Local<Value>`.  Is that just getting implicitly cast to `MaybeLocal<Value>`?  If so, then we should get rid of that implicit cast to make the code clearer and avoid unnecessary error checking.

genevieve: I think that's the pattern of these functions for anything that can return Local be cast to MaybeLocal and be checked for emptiness so the `RtnValue` can get an error instead of the other fields set.

dylanahsmith: It wouldn't ever be empty if v8 returns a Local instead of a MaybeLocal, so the Function.SourceMapUrl should reflect this too.  Any error checking introduced to handle this cast is just dead code that adds confusion all the way up to the code using the library.

v8go.cc
```diff
@@ -1207,6 +1207,19 @@ RtnValue FunctionNewInstance(ValuePtr ptr, int argc, ValuePtr args[]) {
...
+  RtnValue rtn = {nullptr, nullptr};
+  Local<Function> fn = Local<Function>::Cast(value);
+  Local<Value> result = fn->GetScriptOrigin().SourceMapUrl();
+  m_value* rtnval = new m_value;
+  rtnval->iso = iso;
+  rtnval->ctx = ctx;
+  rtnval->ptr = Persistent<Value, CopyablePersistentTraits<Value>>(iso, result);
+  rtn.value = tracked_value(ctx, rtnval);
+  return rtn;
+}
```
dylanahsmith, line 1221: The return value is still implying that this could return an error, but there isn't a code path that would return an error.

```suggestion
ValuePtr FunctionSourceMapUrl(ValuePtr ptr) {
  LOCAL_VALUE(ptr)
  RtnValue rtn = {nullptr, nullptr};
  Local<Function> fn = Local<Function>::Cast(value);
  Local<Value> result = fn->GetScriptOrigin().SourceMapUrl();
  m_value* rtnval = new m_value;
  rtnval->iso = iso;
  rtnval->ctx = ctx;
  rtnval->ptr = Persistent<Value, CopyablePersistentTraits<Value>>(iso, result);
  return tracked_value(ctx, rtnval);
}
```

v8go.h
```diff
@@ -180,6 +180,7 @@ extern ValuePtr PromiseResult(ValuePtr ptr);
 
 extern RtnValue FunctionCall(ValuePtr ptr, int argc, ValuePtr argv[]);
 RtnValue FunctionNewInstance(ValuePtr ptr, int argc, ValuePtr args[]);
+RtnValue FunctionSourceMapUrl(ValuePtr ptr);
```
dylanahsmith, line 183:
```suggestion
ValuePtr FunctionSourceMapUrl(ValuePtr ptr);
```

function.go
```diff
@@ -46,3 +46,9 @@ func (fn *Function) NewInstance(args ...Valuer) (*Object, error) {
 	fn.ctx.deregister()
 	return getObject(fn.ctx, rtn), getError(rtn)
 }
+
+// Return the script origin for a function.
+func (fn *Function) SourceMapUrl() *Value {
+	rtn := C.FunctionSourceMapUrl(fn.ptr)
+	return getValue(fn.ctx, rtn)
```
dylanahsmith, line 53:
```suggestion
	ptr := C.FunctionSourceMapUrl(fn.ptr)
	return &Value{ptr, fn.ctx}
```

function.go
```diff
@@ -46,3 +46,9 @@ func (fn *Function) NewInstance(args ...Valuer) (*Object, error) {
 	fn.ctx.deregister()
 	return getObject(fn.ctx, rtn), getError(rtn)
 }
+
+// Return the script origin for a function.
```
dylanahsmith, line 50: This doesn't return the script origin, it returns the source map url.  It also exposes it directly on the Function type instead of through a ScriptOrigin struct.  Is there a reason why you chose to not expose a ScriptOrigin type to more closely match the original API?  Especially given that matching the original API is a project goal unless it conflicts with matching what Gophers would expect.

genevieve: Mostly introducing the whole new type to the v8go codebase and needing to figure out memory management around it led me to this implementation but open to exposing the script origin type instead and giving it a source map url getter.

dylanahsmith: That seems like a good reason to keep it the way it is right now.

This documentation comment could still use updating though.

dylanahsmith (approved):

PR #149, opened 2021-06-25
refactor: Extract func coerceValue from set and remove func set

dylanahsmith:
cc @wisepythagoras

## Problem

https://github.com/rogchap/v8go/pull/145 is attempting to follow the current pattern of merging more code paths into `set`, which results in extra conditional complexity in `set` in order to essentially determine where it is called from and thus which argument to use.

## Solution

The common functionality in the `set` function is value coercion, so just extract that to a separate function so that we can get rid of `set`.

Note that this makes all of these functions (coerceValue, Set, SetIdx) a lot more readable in isolation.  In contrast, the call site for `set` makes it unclear why the placeholder arguments are passed in and the `set` function is harder to understand without knowing the context in which it is called from.

### DIFF
--- a/object.go
+++ b/object.go
@@ -47,6 +47,20 @@ func (o *Object) MethodCall(methodName string, args ...Valuer) (*Value, error) {
 	return getValue(fn.ctx, rtn), getError(rtn)
 }
 
+func coerceValue(iso *Isolate, val interface{}) (*Value, error) {
+	switch v := val.(type) {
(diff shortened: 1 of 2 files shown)


wisepythagoras:
I like this solution. I'll incorporate it in #145 once it's merged.

wisepythagoras (approved):

rogchap (approved):
Reads better, thank you.

PR #152, opened 2021-07-06
Avoid error in type signature for Context.Isolate, NewObjectTemplate & NewFunctionTemplate

### LINKED ISSUE #108: Consider using panic
Every method in the API returns an error, which makes it unnecessarily cumbersome to use. 

For example

```go
// NewFunctionTemplate creates a FunctionTemplate for a given callback.
func NewFunctionTemplate(iso *Isolate, callback FunctionCallback) (*FunctionTemplate, error) {
	if iso == nil {
		return nil, errors.New("v8go: failed to create new FunctionTemplate: Isolate cannot be <nil>")
	}
	if callback == nil {
		return nil, errors.New("v8go: failed to create new FunctionTemplate: FunctionCallback cannot be <nil>")
	}

	cbref := iso.registerCallback(callback)

	tmpl := &template{
		ptr: C.NewFunctionTemplate(iso.ptr, C.int(cbref)),
		iso: iso,
	}
	runtime.SetFinalizer(tmpl, (*template).finalizer)
	return &FunctionTemplate{tmpl}, nil
}
```

In this case, I would argue that it should panic if either of the parameter are nil, because this is incorrect use of the API and is trivially able to be checked by a caller who might have an Isolate or FunctionCallback of unknown provenance. 

To draw a comparison, `strings.Repeat` panics if an invalid argument is provided, and I have never heard of that being considered problematic. Instead, returning many errors that can never happen either results in callers needing to do one of 3 things:

1. Add a panic in their code, in which case, what has been saved? This is what I do, unhappily. 
2. Ignore the error. This is bad, because they would need to view the implementation to know which errors are impossible.
3. Handle the error. This is time wasted and additional code to maintain

Obviously it's a breaking change, but I think it would be a welcome one to remove error returns of this sort from the API. 

Thanks for considering.

dylanahsmith:
Fixes #108

The above issue linked to https://github.com/rogchap/v8go/pull/118#discussion_r619480410 where it was decided that `panic` may be more appropriate and idiomatic Go code in certain cases

Specifically, this makes sense when:
* there is no underlying error in the v8 API used
* any error can be equivalently checked in the caller
* the error typically represents a bug in the caller

As such, I removed error from the type signature of:
* Context.Isolate
* NewObjectTemplate
* NewFunctionTemplate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for calling constructors functions with NewInstance on Function
 - Build v8 with i18n support
 
+### Changed
+- Removed error return value from Context.Isolate() which never fails
+- Removed error return value from NewObjectTemplate and NewFunctionTemplate. Panic if given a nil argument.
+
 ## [v0.6.0] - 2021-05-11
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ case val := <- vals:
 case err := <- errs:
     // javascript error
 case <- time.After(200 * time.Milliseconds):
-    vm, _ := ctx.Isolate() // get the Isolate from the context
+    vm := ctx.Isolate() // get the Isolate from the context
     vm.TerminateExecution() // terminate the execution 
     err := <- errs // will get a termination error back from the running script
 }
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ printfn := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *
     fmt.Printf("%v", info.Args()) // when the JS function is called this Go callback will execute
     return nil // you can return a value back to the JS caller if required
 })
-global, _ := v8go.NewObjectTemplate(iso) // a template that represents a JS Object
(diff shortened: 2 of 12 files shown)


dylanahsmith:
An example that highlights the need for this change is that code using NewContext to create the isolate (e.g. `ctx, err := v8go.NewContext()`).  If that succeeds, the code would be expected to close the context and dispose of the isolate to avoid a memory leak.  However, the error return value prevents the isolate from being disposed of unconditionally using `defer ctx.Isolate().Dispose()`.

wisepythagoras (approved):

wisepythagoras:
@dylanahsmith It looks like @rogchap doesn't really care about this project anymore, so in the coming days I'll be pulling all your PRs into my forked branch to - hopefully - continue development there.

dylanahsmith:
I talked to @rogchap on the v8go slack channel and it isn't that he doesn't care about this project, he has just been preoccupied.

For now, I would recommend continuing to contribute changes to this repo and using a fork to pull in any features blocking your work.  That is the approach we are taking.  In the case of this PR, it isn't a feature that is needed to unblock anything.

wisepythagoras:
Thanks for the clarification. I didn't realize that there was a Slack channel about this. Sounds good!

genevieve (approved):

rogchap (approved):

Issue #156, opened 2021-08-24
Calling an instance method that uses `this` fails with `undefined`


genevieve:
I have a Response object with a constructor that accepts a body as argument, and has a method called `text()` that returns said body as string. When I get the response as an object and call the method, I get `Cannot read property 'body' of undefined`.

Here's the example go code:
```go
package main

import (
	"fmt"
	"io/ioutil"

	"rogchap.com/v8go"
)

func main() {
	script, _ := ioutil.ReadFile("script.js")
	iso, _ := v8go.NewIsolate() // creates a new JavaScript VM
	ctx, _ := v8go.NewContext(iso)
	val, _ := ctx.RunScript(string(script), "script.js")
	respObj, _ := val.AsObject()
	textVal, _ := respObj.Get("text")
	textFn, _ := textVal.AsFunction()
	body, err := textFn.Call()
	if err != nil {
		panic(err)
	}
	fmt.Println(body)
}
```

Here's the script:
```js
class Response {
  constructor(body) {
    this.body = body;
  }
  text() {
    return this.body.toString();
  }
}

new Response("this body");
```

Here's the error:
```
panic: TypeError: Cannot read property 'body' of undefined

goroutine 1 [running]:
main.main()
	/Users/genevieve/workspace/test/main.go:35 +0x37c
exit status 2
```

I'm assuming I'm doing something wrong but not sure exactly what it is. Any guidance is appreciated 🙏🏻 

EDIT: There might be parts of https://github.com/rogchap/v8go/pull/142/files that are relevant.

EDIT 2: Questions - Should a v8go function support `bind()`? Should a v8go function keep a reference to the object it belongs to with an internal `this` ptr? Is this function not really executing directly on the instance of response that I think it is?

dylanahsmith:
What you are doing from Go corresponds to javascript like

```
r = new Response("this body")
f = r.text
f()
```

which gets the same error, since a function object doesn't carry a receiver with it.

The implicit passing of `undefined` as the receiver seems like an unfortunate gotcha to the javascript language, which more experienced javascript developers may be familiar with, but can easily cause confusion when first encountered.

v8's API more closely follows the [Function.prototype.call javascript method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) in passing the receiver explicitly, but actually requires the receiver a required parameter.  This avoids the confusion of implicitly passing `undefined`.

[...]

PR #159, opened 2021-09-01
Function Call should support recv argument

genevieve:
(cherry picked from commit 991ab091d967512721296e380a731efa08c9f85c)

Fixes https://github.com/rogchap/v8go/issues/156

The v8go library is implicitly passing `undefined` as the receiver when executing a function `Call`. To make this configurable, as shown in this PR, it would be a breaking change to the v8go api for the function object. This change ensures the explicit passing of a receiver to a function and avoid additional confusion around the fact that functions obtained from an object are not bound to the object they are obtained from. If we want to avoid the breaking change, we could instead expose something on the v8go object type called `MethodCall` which would ensure the proper execution a function/method with the object as it's receiver and avoid code like:
```
respVal, _ := ctx.Get("response")
respObj, _ := respVal.AsObject()
textVal, _ := respObj.Get("text")
textFn, _ := textVal.AsFunction()
textProm, _ := textFn.Call(respObj)
...
```

And instead may look like:
```
respVal, _ := ctx.Get("response")
respObj, _ := respVal.AsObject()
textProm, _ := respObj.MethodCall("text")
...
```

The choice is essentially - a breaking change that aligns the function call with the v8 api _or_ deviating a little from the v8 api by exposing a new method on objects.

### DIFF
--- a/function.go
+++ b/function.go
@@ -16,7 +16,7 @@ type Function struct {
 }
 
 // Call this JavaScript function with the given arguments.
-func (fn *Function) Call(recv *Value, args ...Valuer) (*Value, error) {
+func (fn *Function) Call(recv Valuer, args ...Valuer) (*Value, error) {
 	var argptr *C.ValuePtr
 	if len(args) > 0 {
 		var cArgs = make([]C.ValuePtr, len(args))
@@ -26,7 +26,7 @@ func (fn *Function) Call(recv *Value, args ...Valuer) (*Value, error) {
 		argptr = (*C.ValuePtr)(unsafe.Pointer(&cArgs[0]))
 	}
 	fn.ctx.register()
-	rtn := C.FunctionCall(fn.ptr, recv.ptr, C.int(len(args)), argptr)
+	rtn := C.FunctionCall(fn.ptr, recv.value().ptr, C.int(len(args)), argptr)
 	fn.ctx.deregister()
 	return getValue(fn.ctx, rtn), getError(rtn)
 }
--- a/function.go
+++ b/function.go
@@ -16,7 +16,7 @@ type Function struct {
 }
 
 // Call this JavaScript function with the given arguments.
-func (fn *Function) Call(args ...Valuer) (*Value, error) {
+func (fn *Function) Call(recv *Object, args ...Valuer) (*Value, error) {
 	var argptr *C.ValuePtr
 	if len(args) > 0 {
 		var cArgs = make([]C.ValuePtr, len(args))
@@ -26,7 +26,7 @@ func (fn *Function) Call(args ...Valuer) (*Value, error) {
 		argptr = (*C.ValuePtr)(unsafe.Pointer(&cArgs[0]))
 	}
 	fn.ctx.register()
-	rtn := C.FunctionCall(fn.ptr, fn.ctx.iso.undefined.ptr, C.int(len(args)), argptr)
+	rtn := C.FunctionCall(fn.ptr, recv.ptr, C.int(len(args)), argptr)
 	fn.ctx.deregister()
 	return getValue(fn.ctx, rtn), getError(rtn)
 }
--- a/function.go
+++ b/function.go
@@ -16,7 +16,7 @@ type Function struct {
 }
 
 // Call this JavaScript function with the given arguments.
-func (fn *Function) Call(recv *Object, args ...Valuer) (*Value, error) {
+func (fn *Function) Call(recv *Value, args ...Valuer) (*Value, error) {
 	var argptr *C.ValuePtr
 	if len(args) > 0 {
 		var cArgs = make([]C.ValuePtr, len(args))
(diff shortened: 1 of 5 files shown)


dylanahsmith:
> If we want to avoid the breaking change, we could instead expose something on the v8go object type called `MethodCall` which would ensure the proper execution a function/method with the object as it's receiver

As I mentioned in https://github.com/rogchap/v8go/issues/156#issuecomment-909427012, I think that would be a useful convenience method even with the breaking change.  It would also reduce the need for using Function.Call directly.

Making the receiver explicit in Function.Call is independently useful in making Function.Call more generally useful when passing an explicit receiver is necessary, to avoiding the confusion that comes from implicitly passing `undefined` and to keep the API more consistent with the v8 API.

genevieve:
Ok I will leave this PR here and we can merge it if passes review and we plan to ship a release of v8go with the breaking change. I can make a separate PR for the new object.methodcall suggested.

function.go
```diff
@@ -16,7 +16,7 @@ type Function struct {
 }
 
 // Call this JavaScript function with the given arguments.
-func (fn *Function) Call(args ...Valuer) (*Value, error) {
+func (fn *Function) Call(recv *Object, args ...Valuer) (*Value, error) {
```
dylanahsmith, line 19: [v8::Function::Call](https://v8.github.io/api/head/classv8_1_1Function.html#ad916775f129a95ed93e4c68cd3848b8b) uses `Local<Value>` as the receiver type, which we should do for parity

```suggestion
func (fn *Function) Call(recv *Value, args ...Valuer) (*Value, error) {
```

as well as to support passing in an undefined value for top-level function calls, which probably makes more sense when updating existing calls

```diff
diff --git a/function_template_test.go b/function_template_test.go
index b2fed39..1f9dca1 100644
--- a/function_template_test.go
+++ b/function_template_test.go
@@ -71,7 +71,7 @@ func TestFunctionTemplateGetFunction(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	ret, err := fn.Call(ctx.Global(), ten)
+	ret, err := fn.Call(v8go.Undefined(iso), ten)
 	if err != nil {
 		t.Fatal(err)
 	}
diff --git a/function_test.go b/function_test.go
index 4bf1040..6c635f6 100644
--- a/function_test.go
+++ b/function_test.go
@@ -28,7 +28,7 @@ func TestFunctionCall(t *testing.T) {
 	failIf(t, err)
 
 	fn, _ := addValue.AsFunction()
-	resultValue, err := fn.Call(ctx.Global(), arg1, arg1)
+	resultValue, err := fn.Call(v8go.Undefined(iso), arg1, arg1)
 	failIf(t, err)
 
 	if resultValue.Int32() != 2 {
@@ -91,7 +91,7 @@ func TestFunctionCallToGoFunc(t *testing.T) {
 	failIf(t, err)
 	fn, err := val.AsFunction()
 	failIf(t, err)
-	resultValue, err := fn.Call(ctx.Global())
+	resultValue, err := fn.Call(v8go.Undefined(iso))
 	failIf(t, err)
 
 	if !called {
@@ -107,7 +107,8 @@ func TestFunctionCallError(t *testing.T) {
 
 	ctx, err := v8go.NewContext()
 	failIf(t, err)
-	defer ctx.Isolate().Dispose()
+	iso := ctx.Isolate()
+	defer iso.Dispose()
 	defer ctx.Close()
 
 	_, err = ctx.RunScript("function throws() { throw 'error'; }", "script.js")
@@ -116,7 +117,7 @@ func TestFunctionCallError(t *testing.T) {
 	failIf(t, err)
 
 	fn, _ := addValue.AsFunction()
-	_, err = fn.Call(ctx.Global())
+	_, err = fn.Call(v8go.Undefined(iso))
 	if err == nil {
 		t.Errorf("expected an error, got none")
 	}
```

dylanahsmith (approved):

function.go
```diff
@@ -16,7 +16,7 @@ type Function struct {
 }
 
 // Call this JavaScript function with the given arguments.
-func (fn *Function) Call(args ...Valuer) (*Value, error) {
+func (fn *Function) Call(recv *Value, args ...Valuer) (*Value, error) {
```
rogchap, line 19: Would the receiver not be a `Valuer` too?

rogchap (approved):
One small comment; otherwise looks good.

I'm ok with the breaking change as we are on 0.x and we have already made some breaking changes with using panic rather than returning null errors.
Would be good to document some of these breaking changes in the Readme.

dylanahsmith:
> Would be good to document some of these breaking changes in the Readme.

They are being documented in the CHANGELOG.md and when a release is made, we could copy that text to the release description.  Those seem like the places where someone would go to find information on changes relevant to an upgrade.

Probably the main point of confusion with the README is that someone encountering the project for the first time will see the README for the `master` branch rather than for the latest release, which is a common problem with reading it on github.  At least it will show the right thing when viewing the documentation (https://pkg.go.dev/rogchap.com/v8go#section-readme), so perhaps we could refer them for the latest release documentation if there are unreleased breaking changes that affect the README.

dylanahsmith:
Alternatively, perhaps it would be a good time to make a new release, which would also avoid the confusion from unreleased breaking changes.

rogchap:
I think this is a good idea; a new release is overdue.

PR #160, opened 2021-09-01
Support calling a method on an object.

### LINKED ISSUE #156: Calling an instance method that uses `this` fails with `undefined`

genevieve:
An extension of #159 to fix #156

https://github.com/rogchap/v8go/pull/159#issuecomment-910678699

Builds on #162

### DIFF
--- a/function.go
+++ b/function.go
@@ -26,7 +26,7 @@ func (fn *Function) Call(args ...Valuer) (*Value, error) {
 		argptr = (*C.ValuePtr)(unsafe.Pointer(&cArgs[0]))
 	}
 	fn.ctx.register()
-	rtn := C.FunctionCall(fn.ptr, nil, C.int(len(args)), argptr)
+	rtn := C.FunctionCall(fn.ptr, fn.ctx.iso.undefined.ptr, C.int(len(args)), argptr)
 	fn.ctx.deregister()
 	return getValue(fn.ctx, rtn), getError(rtn)
 }
--- a/function.go
+++ b/function.go
@@ -26,7 +26,7 @@ func (fn *Function) Call(args ...Valuer) (*Value, error) {
 		argptr = (*C.ValuePtr)(unsafe.Pointer(&cArgs[0]))
 	}
 	fn.ctx.register()
-	rtn := C.FunctionCall(fn.ptr, C.int(len(args)), argptr)
+	rtn := C.FunctionCall(fn.ptr, nil, C.int(len(args)), argptr)
 	fn.ctx.deregister()
 	return getValue(fn.ctx, rtn), getError(rtn)
 }
--- a/object.go
+++ b/object.go
@@ -28,9 +28,7 @@ func (o *Object) MethodCall(methodName string, args ...Valuer) (*Value, error) {
 	if err != nil {
 		return nil, err
 	}
-	fnVal := getValue(o.ctx, getRtn)
-
-	fn, err := fnVal.AsFunction()
+	fn, err := getValue(o.ctx, getRtn).AsFunction()
 	if err != nil {
 		return nil, err
 	}
--- a/object.go
+++ b/object.go
@@ -19,6 +19,36 @@ type Object struct {
 	*Value
 }
 
+func (o *Object) MethodCall(methodName string, args ...Valuer) (*Value, error) {
+	ckey := C.CString(methodName)
+	defer C.free(unsafe.Pointer(ckey))
+
+	getRtn := C.ObjectGet(o.ptr, ckey)
+	err := getError(getRtn)
+	if err != nil {
+		return nil, err
+	}
+	fnVal := getValue(o.ctx, getRtn)
+
+	fn, err := fnVal.AsFunction()
+	if err != nil {
+		return nil, err
+	}
+
+	var argptr *C.ValuePtr
+	if len(args) > 0 {
+		var cArgs = make([]C.ValuePtr, len(args))
+		for i, arg := range args {
+			cArgs[i] = arg.value().ptr
+		}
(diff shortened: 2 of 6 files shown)


object.go
```diff
@@ -19,6 +19,36 @@ type Object struct {
...
+		var cArgs = make([]C.ValuePtr, len(args))
+		for i, arg := range args {
+			cArgs[i] = arg.value().ptr
+		}
+		argptr = (*C.ValuePtr)(unsafe.Pointer(&cArgs[0]))
+	}
+	fn.ctx.register()
+	rtn := C.FunctionCall(fn.ptr, o.ptr, C.int(len(args)), argptr)
+	fn.ctx.deregister()
+	return getValue(fn.ctx, rtn), getError(rtn)
```
genevieve, line 49: Lines 38-49 can be replaced with `fn.Call(obj, args...)` once the other PR is merged. I kept them separate since one is a breaking change.

v8go.cc
```diff
@@ -1168,14 +1168,24 @@ static void buildCallArguments(Isolate* iso, Local<Value> *argv, int argc, Value
...
   Local<Function> fn = Local<Function>::Cast(value);
   Local<Value> argv[argc];
   buildCallArguments(iso, argv, argc, args);
-  Local<Value> recv = Undefined(iso);
-  MaybeLocal<Value> result = fn->Call(local_ctx, recv, argc, argv);
+
+  MaybeLocal<Value> result;
+
+  if (recv == NULL) {
+    result = fn->Call(local_ctx, Undefined(iso), argc, argv);
```
genevieve, line 1182: To maintain backwards compatibility for the API but actually correctly implement the `FunctionCall` for the v8 api, I was thinking of supporting either a nul or obj receiver and handling either case internally so the v8go function method does not need to change as part of this PR.

v8go.cc
```diff
@@ -1168,14 +1168,24 @@ static void buildCallArguments(Isolate* iso, Local<Value> *argv, int argc, Value
...
   RtnValue rtn = {nullptr, nullptr};
   Local<Function> fn = Local<Function>::Cast(value);
   Local<Value> argv[argc];
   buildCallArguments(iso, argv, argc, args);
-  Local<Value> recv = Undefined(iso);
-  MaybeLocal<Value> result = fn->Call(local_ctx, recv, argc, argv);
+
+  MaybeLocal<Value> result;
+
+  if (recv == NULL) {
```
dylanahsmith, line 1181: This feels like an unnecessary conditional, since it is just distinguishing between different call sites.  That means we can remove the conditional by moving code to the caller.

I've opened https://github.com/rogchap/v8go/pull/163 to allow the Go code to explicitly pass an undefined value, so it can be treated like any other value here.

function.go
```diff
@@ -26,7 +26,7 @@ func (fn *Function) Call(args ...Valuer) (*Value, error) {
 		argptr = (*C.ValuePtr)(unsafe.Pointer(&cArgs[0]))
 	}
 	fn.ctx.register()
-	rtn := C.FunctionCall(fn.ptr, C.int(len(args)), argptr)
+	rtn := C.FunctionCall(fn.ptr, fn.ctx.iso.undefined.ptr, C.int(len(args)), argptr)
```
genevieve, line 29: A little 😬

object_test.go
```diff
@@ -11,6 +11,35 @@ import (
...
+	if err == nil {
+		t.Errorf("expected an error, got none")
+	}
+
+	val, err = ctx.RunScript(`class Obj2 { print(str) { return str.toString() }; get fails() { throw "error" } }; new Obj2()`, "")
+	failIf(t, err)
+	obj, _ = val.AsObject()
+	arg, _ := v8go.NewValue(iso, "arg")
+	_, err = obj.MethodCall("print", arg)
+	failIf(t, err)
```
dylanahsmith, line 36: We should probably actually check that the right value got passed in by checking the return value.  I'll push a commit that does this

```suggestion
	val, err = obj.MethodCall("print", arg)
	failIf(t, err)
	if val.String() != "arg" {
		t.Errorf("unexpected value: %q", val)
	}
```

dylanahsmith (approved):

PR #162, opened 2021-09-03
Add value.SameValue(otherValue) function

dylanahsmith:
When comparing objects, it would be useful to be able to test that the objects are the same and don't just have the same data, so I exposed v8::Value::SameValue


genevieve (approved):

rogchap (approved):
Looks all good.

PR #163, opened 2021-09-03
Expose null and undefined constant values

dylanahsmith:
Builds on #162 
cc @genevieve 

Currently, there is no simple way of getting `null` or `undefined` JS values from v8go.  There are workarounds like running a script to get these values, but this shouldn't be necessary.

For instance, this would be useful in #160 for explicitly passing the `undefined` value as an explicit function call receiver


genevieve (approved):

rogchap (approved):
Makes sense; thank you.

PR #166, opened 2021-09-07
Include line number information in debug builds of v8

dylanahsmith:
cc @genevieve 

## Problem

While @ofuerst was looking into adding support for v8::ScriptCompiler::CompileUnboundScript, he encountered crashes which weren't easy to debug using the pre-build v8go libv8.  I tried to build v8 with debugging information to get a C++ backtrace to help debug the problem, although this involved modifying deps/build.py and also wasn't very clear to someone not already used to debugging crashes from cgo integrations.  Note that the problem ended up being caught by V8 debug checking.

## Solution

I've modified deps/build.py to include line number information when that script is used with the `--debug` flag.

I've also documented the steps I used get a C++ backtrace using a debugger.  I tried out the steps on MacOS with lldb and Ubuntu with both lldb and gdb.  Although, I found that `-ldflags=-compressdwarf=false` was needed on MacOS with lldb, but wasn't needed on Ubuntu for either of these debuggers.

I tried compiling libv8 with different `symbol_level`s and got the following for the deps/darwin_x86_64  sizes:
* 32 MiB for `symbol_level=0`
* 206 MiB `symbol_level=1`
* 3.4 GiB `symbol_level=2`
where that dramatic increase in library sizes had a significant effect on compilation of the v8go test suite (with `go test -c`), although the test suite executable does get smaller, which I assume is due to dead code elimination.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,15 @@ This project also aims to keep up-to-date with the latest (stable) release of V8
 
 ## Development
 
+### Recompile V8 with debug info and debug checks
+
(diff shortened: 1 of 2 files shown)


genevieve (approved):

tmc (approved):
LGTM

ofuerst:
Awesome! Thanks @dylanahsmith

PR #167, opened 2021-09-08
Introduce CpuProfiler, CpuProfile, and CpuProfileNode

genevieve:
https://v8docs.nodesource.com/node-16.0/d2/d34/classv8_1_1_cpu_profiler.html
https://v8docs.nodesource.com/node-16.0/db/d79/classv8_1_1_cpu_profile.html
https://v8docs.nodesource.com/node-16.0/d9/d6e/classv8_1_1_cpu_profile_node.html

There are plenty more functions on the respective classes to implement, but this is a start.

### DIFF
--- a/cpuprofilenode.go
+++ b/cpuprofilenode.go
@@ -1,0 +1,57 @@
+// Copyright 2019 Roger Chapman and the v8go contributors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package v8go
+
+// #include <stdlib.h>
+// #include "v8go.h"
+import "C"
+
+// CPUProfileNode represents a node in a call graph.
+type CPUProfileNode struct {
+	ptr C.CpuProfileNodePtr
+	iso *Isolate
+}
+
+// Returns function name (empty string for anonymous functions.)
+func (c *CPUProfileNode) GetFunctionName() string {
+	str := C.CpuProfileNodeGetFunctionName(c.ptr)
+	return C.GoString(str)
+}
+
+// Retrieves number of children.
+func (c *CPUProfileNode) GetChildrenCount() int {
+	i := C.CpuProfileNodeGetChildrenCount(c.ptr)
+	return int(i)
+}
+
+// Retrieves a child node by index.
+func (c *CPUProfileNode) GetChild(index int) *CPUProfileNode {
+	count := c.GetChildrenCount()
+	if index < 0 || index > count {
+		return nil
+	}
+	ptr := C.CpuProfileNodeGetChild(c.ptr, C.int(index))
+	return &CPUProfileNode{ptr: ptr, iso: c.iso}
+}
+
+// Retrieves the ancestor node, or null if the root.
+func (c *CPUProfileNode) GetParent() *CPUProfileNode {
+	ptr := C.CpuProfileNodeGetParent(c.ptr)
+	return &CPUProfileNode{ptr: ptr, iso: c.iso}
+}
+
+// Returns the number, 1-based, of the line where the function originates.
+// kNoLineNumberInfo if no line number information is available.
+func (c *CPUProfileNode) GetLineNumber() int {
+	i := C.CpuProfileNodeGetLineNumber(c.ptr)
+	return int(i)
+}
+
+//  Returns 1-based number of the column where the function originates.
+//  kNoColumnNumberInfo if no column number information is available.
+func (c *CPUProfileNode) GetColumnNumber() int {
+	i := C.CpuProfileNodeGetColumnNumber(c.ptr)
+	return int(i)
+}
--- a/cpuprofiler.go
+++ b/cpuprofiler.go
@@ -11,94 +11,20 @@ import (
 	"unsafe"
 )
 
-type CpuProfiler struct {
+type CPUProfiler struct {
 	ptr C.CpuProfilerPtr
 	iso *Isolate
 }
 
-// CpuProfile contains a CPU profile in a form of top-down call tree
-// (from main() down to functions that do all the work).
-type CpuProfile struct {
-	ptr C.CpuProfilePtr
-	iso *Isolate
-}
-
-// CpuProfileNode represents a node in a call graph.
-type CpuProfileNode struct {
-	ptr C.CpuProfileNodePtr
-	iso *Isolate
-}
-
-// Returns the root node of the top down call tree.
-func (c *CpuProfile) GetTopDownRoot() *CpuProfileNode {
-	ptr := C.CpuProfileGetTopDownRoot(c.ptr)
-	return &CpuProfileNode{ptr: ptr, iso: c.iso}
-}
-
-// Returns function name (empty string for anonymous functions.)
-func (c *CpuProfileNode) GetFunctionName() string {
-	str := C.CpuProfileNodeGetFunctionName(c.ptr)
-	return C.GoString(str)
-}
-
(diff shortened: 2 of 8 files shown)


cpuprofiler.go
```diff
@@ -0,0 +1,46 @@
...
+// Dispose will dispose the profiler; subsequent calls will panic.
+func (c *CPUProfiler) Dispose() {
+	if c.ptr == nil {
+		return
+	}
+	C.CpuProfilerDispose(c.ptr)
+	c.ptr = nil
+}
+
+func (c *CPUProfiler) StartProfiling(title string) {
```
tmc, line 35: can you include method comments here?

tmc (commented):
Mostly LGTM, one small request.

tmc:
Can you open an issue to add documentation around how to use these exposed types?

tmc (commented):
Hmm, I wonder if this is a good time to consider introducing some additional packages to expose different aspects of v8's api. I imagine most users won't leverage these types and I wonder if they should be placed in a different package.

@rogchap thoughts?

dylanahsmith:
> Having parity with the v8 namespace is certainly a nice quality but as we add more surface area it will make use of the package for newcomers more complicated. (I imagine most users won't initially leverage these types).

For this users, is this mostly a documentation problem?  Note that this could easily introduce the opposite problem of making it harder for users to find functionality when it isn't clear which package it is in.

> mentioned that we could follow how v8 [splits up header files](https://github.com/v8/v8/tree/master/include) (e.g. profiler, inspector).

Note that there are 17 top-level header files in the version of v8 that v8go is currently using and v8 master now has 
59, so I wouldn't follow that file structure too carefully.  For instance, I would treat all the headers included by v8.h as if they are part of v8.h, which would also keep the correspondence fairly stable by coupling to V8's intentionally public API.

In order to split the packages, we might also need an internal/unsafe package that could expose an V8 pointers that need to be accessed from other packages.

cpuprofiler.go
```diff
@@ -0,0 +1,54 @@
...
+	c.ptr = nil
+}
+
+// StartProfiling starts collecting a CPU profile. Title may be an empty string. Several
+// profiles may be collected at once. Attempts to start collecting several
+// profiles with the same title are silently ignored.
+func (c *CPUProfiler) StartProfiling(title string) {
+	tstr := C.CString(title)
+	defer C.free(unsafe.Pointer(tstr))
+	C.CpuProfilerStartProfiling(c.iso.ptr, c.ptr, tstr)
```
dylanahsmith, line 42: Documentation for CPUProfiler.Dispose says "subsequent calls will panic", but there are no corresponding checks for `c.iso.ptr` or `c.ptr` being `nil` to panic if either has been disposed.

cpuprofilenode.go
```diff
@@ -0,0 +1,55 @@
...
+package v8go
+
+// #include <stdlib.h>
+// #include "v8go.h"
+import "C"
+
+// CPUProfileNode represents a node in a call graph.
+type CPUProfileNode struct {
+	ptr C.CpuProfileNodePtr
+	iso *Isolate
```
dylanahsmith, line 14: At what point is the `v8::CpuProfileNode`'s memory freed?  Is that when the profiler is freed or does `v8::CpuProfile::Delete` need to be called?  Either way, it seems like we need to be able to panic to avoid undefined behaviour if the `v8::CpuProfileNode` is used after its memory is freed.

cpuprofilenode.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Returns function name (empty string for anonymous functions.)
+func (c *CPUProfileNode) GetFunctionName() string {
```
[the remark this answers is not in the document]
genevieve: I thought the intention was to have the Go API match the V8 API. In this case and others for these components, the v8 functions are called `GetSomething`. It felt unintuitive to not follow that convention since we are trying to map as closely to v8 where we can and make discovering docs for our own functions easy to find in v8 api docs.

https://github.com/rogchap/v8go/blob/ed9a5f0761e2d7ff86dc622fb742f35f570c905e/deps/include/v8-profiler.h#L86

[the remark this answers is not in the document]
dylanahsmith: v8go is both trying to match the v8 API and provide an API that users of go would expect, which why I made the suggestion.  However, it does look like there are a few existing functions that use the `Get` prefix (GetFunction, GetPromise, GetHeapStatistics).  If we were to differ from v8 to make it more idiomatic for Go, we should be consistent about it.  So I think you are right in using the Get prefix

cpuprofiler.go
genevieve: I actually don't know where I got that documentation/comment from. I originally was basing myself on the v8 api docs online then source directly from the header files but neither that I'm looking at right now suggest that this will in fact occur. The `v8-profiler.h` describes `CPUProfiler.Dispose` as `   * Disposes the CPU profiler object.` I'll update the comment to drop that bit unless we prefer the behavior to be a panic in these situations at the go level than when it hits the c++ implementation.

genevieve: Oh I likely examined the behavior on an isolate: https://github.com/rogchap/v8go/blob/10e762dc4017337a7033a09ba3d2a0efa866ce05/isolate.go#L97

In this case, I'm not sure the comment is right either. Is it suggesting that subsequent calls to `Dispose` will panic or subsequent calls on any functions of the isolate will? The former isn't what's actually happening and there's a test to verify that: https://github.com/rogchap/v8go/blob/10e762dc4017337a7033a09ba3d2a0efa866ce05/isolate_test.go#L110

cpuprofilenode.go
genevieve: The documentation for `v8::CpuProfile::Delete` states:
```
   * Deletes the profile and removes it from CpuProfiler's list.
   * All pointers to nodes previously returned become invalid.
```

Are you suggesting that any function on CpuProfileNode in go should panic if the internal ptr to `C.CpuProfileNodePtr` is nil?

cpuprofilenode.go
```diff
@@ -0,0 +1,55 @@
...
+
+// Retrieves the ancestor node, or null if the root.
+func (c *CPUProfileNode) GetParent() *CPUProfileNode {
+	ptr := C.CpuProfileNodeGetParent(c.ptr)
+	return &CPUProfileNode{ptr: ptr, iso: c.iso}
+}
+
+// Returns the number, 1-based, of the line where the function originates.
+func (c *CPUProfileNode) GetLineNumber() int {
+	i := C.CpuProfileNodeGetLineNumber(c.ptr)
```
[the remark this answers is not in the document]
genevieve: I think supporting the basic APIs from V8 for these components would be a good first step to introducing them into the v8go codebase and seeing how they are used by others and ourselves, as well as what optimizations or additional functionality we could provide. I agree that being able to read the entire cpu profile onto an object once profiling is complete would make it easier for a user to retrieve everything off a single go object instead of making multiple calls and incurring the additional Go<->C costs. This however could easily be a separate function or improvement on `StopProfiling` where it loads everything useful onto a Go CpuProfile object.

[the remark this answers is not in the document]
dylanahsmith: If there isn't a use case for reading a partial CPU profile, then perhaps we could just change the implementation internally to read everything in C++ upfront, rather than exposing a separate Go API to the user.  Given that you are exposing data through getter functions, the approach you have should allow for this evolution of the implementation without breaking changes.

PR #169, opened 2021-09-08
Recommend importing as v8 instead of the default of v8go

dylanahsmith:
The `go` suffix makes sense in the context of the repository name, since it otherwise wouldn't be obvious from the url that it refers to Go bindings for v8.  However, when the library is imported, the `go` suffix becomes redundant since it is already in the context of go code.

As mentioned in the Go [Package names](https://go.dev/blog/package-names) blog post

> Good package names are short and clear.

So recommend importing the package as `v8` instead seems like better guidance for users of the library, without require a breaking change.

Note that we may want to consider moving the package to a `v8` subfolder in the repo to instead import using `import "rogchap.com/v8go/v8"`.  However, let's start with the recommendation, which if followed, will only require changes to the import lines if we moved the package to a subfolder like that in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,13 +14,13 @@
 ## Usage
 
 ```go
-import "rogchap.com/v8go"
+import v8 "rogchap.com/v8go"
 ```
 
 ### Running a script
 
 ```go
-ctx := v8go.NewContext() // creates a new V8 context with a new Isolate aka VM
+ctx := v8.NewContext() // creates a new V8 context with a new Isolate aka VM
 ctx.RunScript("const add = (a, b) => a + b", "math.js") // executes a script on the global context
 ctx.RunScript("const result = add(3, 4)", "main.js") // any functions previously added to the context can be called
 val, _ := ctx.RunScript("result", "value.js") // return a value in JavaScript back to Go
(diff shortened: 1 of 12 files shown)


genevieve (approved):

rogchap:
The reasoning behind the go suffix is because of go modules and v2+ versioning.
If we change this to have a `v8` in the import statement would it conflict with Go Modules?
Would `import "rogchap.com/v8go/v8/v2"` still work? better still `import "rogchap.com/v8go/v8/v8"`?
I'm all for "short and clear", but I think that `v8go` is still pretty short and clear even with the redundant suffix.

dylanahsmith:
> The reasoning behind the go suffix is because of go modules and v2+ versioning.

Oh, good point.  It looks like `import "rogchap.com/v8go/v8"` would complain that the version isn't version 8.  `import "rogchap.com/v8go/v8/v2"` would actually work, but go considers the `/v1` or `/v0` suffixes invalid, since it encourages packages to not use a version suffix until the package reaches version 2 and avoids breaking import paths to go from version 0.* to version 1.  So it doesn't look like a `/v8` import path suffix will work.

Technically, the import path could not match the package name, but that wouldn't follow the conventions so would be confusing.

It seems like keeping the package name as v8go and renaming on import makes the most sense for version 0 / 1 of the library.  Technically, it is working around a naming collision, not with another package name, but with import path name suffixes.  Renaming the package still isn't required, it just makes the code using the imported library more readable.

rogchap (approved):

PR #172, opened 2021-09-15
Introduce isolate's IsExecutionTerminating to wait for termination

genevieve:
Introduce isolate's `IsExecutionTerminating()` to check if an execution is still terminating

### DIFF
--- a/isolate_test.go
+++ b/isolate_test.go
@@ -36,7 +36,8 @@ func TestIsolateTermination(t *testing.T) {
 		iso.TerminateExecution()
 	}()
 
-	if e := <-err; e == nil || !strings.HasPrefix(e.Error(), "ExecutionTerminated") {
+	_, e := ctx.RunScript(`while (true) { }`, "forever.js")
+	if e == nil || !strings.HasPrefix(e.Error(), "ExecutionTerminated") {
 		t.Errorf("unexpected error: %v", e)
 	}
 	if iso.IsExecutionTerminating() {
--- a/isolate_test.go
+++ b/isolate_test.go
@@ -27,10 +27,7 @@ func TestIsolateTermination(t *testing.T) {
 
 	var terminating bool
 	fooFn := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *v8go.Value {
-		ctx := info.Context()
-		obj := ctx.Global()
-		val, _ := obj.Get("loop")
-		loop, _ := val.AsFunction()
+		loop, _ := info.Args()[0].AsFunction()
 		loop.Call(v8go.Undefined(iso))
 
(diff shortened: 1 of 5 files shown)


isolate_test.go
```diff
@@ -39,6 +39,9 @@ func TestIsolateTermination(t *testing.T) {
 	if e := <-err; e == nil || !strings.HasPrefix(e.Error(), "ExecutionTerminated") {
 		t.Errorf("unexpected error: %v", e)
 	}
+	if iso.IsExecutionTerminating() {
```
dylanahsmith, line 42: Should we have another test for when this returns `false`?  Should be easy to do that just after creating the isolate.

isolate_test.go
```diff
@@ -39,6 +39,9 @@ func TestIsolateTermination(t *testing.T) {
 	if e := <-err; e == nil || !strings.HasPrefix(e.Error(), "ExecutionTerminated") {
```
dylanahsmith, line 39: By the way, we could move ctx.RunScript out of the unnecessary goroutine it is called in just above this line

```suggestion
	_, e := ctx.RunScript(`while (true) { }`, "forever.js")
	if e == nil || !strings.HasPrefix(e.Error(), "ExecutionTerminated") {
```

isolate_test.go
genevieve: That is checking that it returns false. We'd need a test case to vet it returns true.

genevieve: We'd need to reproduce the situation where terminating execution of some javascript isnt happening as fast as we think it is.

dylanahsmith: In that case, we might need to do the check in a Go callback for a function template.  V8 could terminate nested JS execution, but then would need to return control back to Go code while there would still be a JS frame on the stack.

isolate_test.go
```diff
@@ -20,25 +20,44 @@ func TestIsolateTermination(t *testing.T) {
...
-	go func() {
-		_, e := ctx.RunScript(`while (true) { }`, "forever.js")
-		err <- e
-	}()
+	var terminating bool
+	fooFn := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *v8go.Value {
+		ctx := info.Context()
+		obj := ctx.Global()
+		val, _ := obj.Get("loop")
+		loop, _ := val.AsFunction()
```
dylanahsmith, line 33: Loops like the loop function was passed in to simplify this, then you forgot to actually use it

```suggestion
		loop, _ := info.Args()[0].AsFunction()
```

dylanahsmith (approved):

tmc:
Could you include a PR description here?

genevieve:
@tmc updated

dylanahsmith:
Could you also add a change log entry?

PR #174, opened 2021-09-17
Fix CI on forks by making the FOSSA step conditional on its secret

dylanahsmith:
cc @genevieve 

## Problem

It can be useful to enable Github Actions in a fork to test a branch on CI before opening a PR or without pushing debugging code to the PR branch.  However, when doing that, the "Scan and upload FOSSA data" CI steps will fail because they don't have the corresponding github secret.

The same problem comes up for anyone temporarily maintaining a fork of this repo to unblock themselves while they wait for a change to get accepted and merged in v8go.

## Solution

Make those steps conditional on the secret being provided.  This way nothing will need to be changed in the fork to workaround this problem.


genevieve (approved):
👏🏻

rogchap:
Thanks for this; I thought I fixed this perviously. Guess not.

rogchap (approved):

PR #176, opened 2021-09-21
Reduce unnecessary static casting of pointers received from Go

dylanahsmith:
## Problem

There was a lot of use of void pointers because of the pointer types defined in v8go.h for the C API exposed to Go.  This results in a lot of use of static_cast, which is verbose and is less type safe (since the pointer could just as easily be cast to the wrong pointer type)

## Solution

Forward declare the types in v8go.h without defining them, so Go will treat the pointers as pointers to opaque structs, but they will be pointers to the correct type in v8go.cc.

I couldn't do a forward declaration for `v8::Isolate` when compiling with C, so I used conditional compilation to just forward declare it for c++.  For C, we just want a distinct struct type for the pointer to help avoid mixing pointers.

I also moved `static_cast<m_ctx*>(iso->GetData(0))` to both reduce places where this static cast is done as well as to leverage the name of the function (`isolateInternalContext `) to make the code more readable.

### DIFF
--- a/v8go.cc
+++ b/v8go.cc
@@ -125,8 +125,7 @@ extern "C" {
 
 /********** Isolate **********/
 
-#define ISOLATE_SCOPE(iso_ptr)                   \
-  Isolate* iso = iso_ptr;                        \
+#define ISOLATE_SCOPE(iso)                       \
   Locker locker(iso);                            \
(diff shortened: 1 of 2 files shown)


genevieve (approved):

rogchap (approved):
Nice. This looks a lot cleaner.

PR #177, opened 2021-09-21
Use MaybeLocal ToLocal instead of IsEmpty and ToLocalChecked

dylanahsmith:
## Problem

We want to make sure to propagate V8 errors, which currently isn't being done consistently.  There are still places where MaybeLocal::ToLocalChecked is being used without checking for the error first with MaybeLocal::IsEmpty.  We should make it easier to find where there is missing error handling in v8go.cc.

## Solution

Use MaybeLocal::ToLocal for error handling.  This way the remaining calls to MaybeLocal::ToLocalChecked stand out as places where we are missing error handling.

### DIFF
--- a/v8go.cc
+++ b/v8go.cc
@@ -91,9 +91,9 @@ RtnError ExceptionError(TryCatch& try_catch, Isolate* iso, Local<Context> ctx) {
     rtn.location = CopyString(sb.str());
   }
 
-  MaybeLocal<Value> mstack = try_catch.StackTrace(ctx);
-  if (!mstack.IsEmpty()) {
-    String::Utf8Value stack(iso, mstack.ToLocalChecked());
+  Local<Value> mstack;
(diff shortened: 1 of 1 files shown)


genevieve (approved):
There were a couple other locations that we saw particular issue with `ToLocalChecked` but we can do those as part of a separate PR.

https://github.com/rogchap/v8go/blob/1e8f6b06d4ee48c3ffd47a0d66fdf7845a0f2eff/v8go.cc#L1146-L1182

From the header files, I think these functions also are using the deprecated `Local` return values where the new return is `MaybeLocal`:

https://github.com/rogchap/v8go/blob/8d15ff390bcdb886757ed257a70a99e080a590e9/deps/include/v8.h#L4863-L4866
https://github.com/rogchap/v8go/blob/8d15ff390bcdb886757ed257a70a99e080a590e9/deps/include/v8.h#L4710-L4714

Issue #165, opened 2021-09-07
Automate upgrading v8 binaries


genevieve:
https://github.com/rogchap/v8go#upgrading-the-v8-binaries

> This process is non-trivial, and hopefully we can automate more of this in the future.

We should automate the workflow described above so that it can be at least manually run to upgrade v8 to the latest release that chrome uses. This workflow can be used to keep v8 up-to-date as well as ready for when a CVE is detected.

tmc:
100% in agreement, a github action that carries out these steps would be a great addition.

GustavoCaso:
@genevieve I have been toying on how we could automate this process.

We could create a GitHub action that runs using the [on.schedule](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onschedule) trigger. We can configure based on the [release cycle](https://v8.dev/docs/release-process) of v8. 

The workflow would take care of doing all the steps outlined in the guide reference above and ultimately it would create a pull request with a special name, which [would trigger](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestbranchestags) the already existing [build workflow](https://github.com/rogchap/v8go/blob/master/.github/workflows/v8build.yml)

We would need to change the build workflow from `on: workflow_dispatch` to:

```yml
on:
  pull_request:
    # Sequence of patterns matched against refs/heads
    branches:    
      - 'v8_new_release' # for example
```

So the fully automated process would involve two different workflows.

@genevieve what do you think about this solution? Is there anything that you would like to change?

dylanahsmith:
It seems like the V8 patches (in [deps/windows_x86_64](https://github.com/rogchap/v8go/tree/8d15ff390bcdb886757ed257a70a99e080a590e9/deps/windows_x86_64)) used to build on windows are going to frequently break on V8 upgrades.

genevieve:
@GustavoCaso Yeah the approach sounds good. Last week I was going on the upgrade to v9.3 and wrote a [script](https://github.com/genevieve/v8go/blob/v9_3_upgrade/upgrade-v8.sh) for the steps described in the readme for building the binaries which looks to also be somewhat covered in the build workflow you linked to.

GustavoCaso:
@genevieve @dylanahsmith I've made some progress towards automating the v8 binary upgrade and build.

[...]

PR #181, opened 2021-09-27
Automate v8 build for macosx and linux

GustavoCaso:
Related to https://github.com/rogchap/v8go/issues/165

This PR adds two new workflows: upgrade and automatic build v8 binaries for macOS and Linux.

The upgrade workflow is triggered every first day of the month, and the automatic v8 build gets trigger on every pull_request which branch follows the `v8_<version>_upgrade` pattern.

The upgrade workflow has been tested on my fork. Here is an [example PR](https://github.com/GustavoCaso/v8go/pull/20) that the upgrade workflow opens.


.github/workflows/v8build_not_for_windows.yml
```diff
@@ -0,0 +1,39 @@
+name: Automated V8 Build For MacOSX and Linux
```
dylanahsmith, line 1: I don't think we need another workflow file. We could instead avoid cancelling the whole workflow when one job fails (https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast)

.github/workflows/v8upgrade.yml
```diff
@@ -0,0 +1,48 @@
...
+            - name: Get latest stable commit ref of v8
+              id: v8_latest_commit
+              run: echo ::set-output name=VALUE::$(curl -sS 'https://omahaproxy.appspot.com/all.json?os=linux&channel=stable' | jq '.[0]["versions"][0]["v8_commit"]')
+            - name: Get latest stable version number of v8
+              id: v8_latest_version
+              run: echo ::set-output name=VALUE::$(curl -sS 'https://omahaproxy.appspot.com/all.json?os=linux&channel=stable' | jq '.[0]["versions"][0]["v8_version"]')
+            - name: Checkout v8 to stable commit ref
+              run: cd deps/v8 && git fetch && git checkout ${{steps.v8_latest_commit.outputs.VALUE}}
+            - name: Copy deps/v8/include into deps/include
+              run: cp -r deps/v8/include/* deps/include/
```
rogchap, line 34: Will this remove any files that have been deleted?

Another step that we also need to take is to make sure we have the empty Go files that are imported so that the modules vendor folder works (there is a current problem where this does not work for the `internal` folder). This is a workaround to make sure the header files are copied to the vendor folder when using `go mod vendor`, otherwise folders without Go "code" gets ignored.

This is most tricky if any new folders get added

.github/workflows/v8build_not_for_windows.yml
GustavoCaso: Ohhh that is great. Thanks for linking up the documentation

.github/workflows/v8upgrade.yml
GustavoCaso: @rogchap Thanks for the feedback. It is true that this step is the most tricky one. I'm going to look for ways to double-check all the requirements.

I'm thinking that to check ✅  everything we might as well have a script that would do everything for us and we can invoke it from the Github action, either way, I'm going to work on it and will ping once I have something that works for all cases: files being deleted, and new folders being added which would need to a `vendor.go` file and the extra import in `cgo.go`

GustavoCaso: @rogchap I create a new Python script to take care of updating the include folder https://github.com/rogchap/v8go/pull/181/commits/945bd172813b996593c11a295814df774f05eab8

I have tested the script both without new folders and with new folders and looks like everything works as expected
Is my first writing Python so I'm not sure about best practices or anything so please feel free to point out my various Python flaws or improvements that can be applied.

The logic to update the `cgo.go` file assumes that the import for `_ "rogchap.com/v8go/deps/include"` will never change. From that assumption, I generate the list of imports to add from the directories inside the `deps/include` folder. Is this assumption correct?

GustavoCaso:
@rogchap would you be able to have a look?

.github/workflows/v8upgrade.yml
```diff
@@ -0,0 +1,52 @@
...
+        name: Upgrade V8
+        runs-on: ubuntu-18.04
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v2
+              with:
+                  submodules: true
+                  fetch-depth: 0
+            - name: Check latest v8 stable version
+              run: cd deps && ./check_latest_v8_version.py
```
GustavoCaso, line 18: I'm not sure if using Python for checking the different version is the most optimal. One thing that I wasn't is how to install the `requests` library inside the GitHub action.

Maybe using something like this:

```
- name: Set up Python 3.8
      uses: actions/setup-python@v2
      with:
        python-version: 3.8
    - name: Install dependencies
      run: |
        python -m pip install --upgrade pip
        pip install requests
```

But I'm not sure if it is a little overkill having to pull some python library to do the version check.

I have use Python, because most scripts in the repo are written in Python.

Open to suggestions 😄

dylanahsmith: I know the requests library is popular, but the python standard library has http support.  Looks like the urllib.request and json libraries can be used for this.  For instance, the following worked for me

```python
import urllib.request
import json

CHROME_VERSIONS_URL = 'https://omahaproxy.appspot.com/all.json?os=linux&channel=stable'
with urllib.request.urlopen(CHROME_VERSIONS_URL) as response:
   json_response = response.read()

data = json.loads(json_response)
stable_version = data[0]["versions"][0]["v8_version"]
print(stable_version)
```

deps/check_latest_v8_version.py
```diff
@@ -0,0 +1,31 @@
...
+# Current version
+v8_version_file = open("v8_version", "r")
+current_v8_version_installed = v8_version_file.read().strip()
+
+# Get latest version
+r = requests.get(url = URL)
+
+data = r.json()
+
+latest_stable_v8_version = data[0]["versions"][0]["v8_version"].strip()
```
dylanahsmith, line 28: It doesn't look like there is any padding that needs to be stripped in the JSON v8_version string

```suggestion
latest_stable_v8_version = data[0]["versions"][0]["v8_version"]
```

deps/check_latest_v8_version.py
```diff
@@ -0,0 +1,31 @@
...
+
+result = compare_v8_version(current_v8_version_installed, latest_stable_v8_version)
+sys.exit(result)
```
dylanahsmith, line 31: Is this going to make it seem like the github action failed just because there is nothing to do?  It seems like that would make it harder to tell when the github action is actually broken

.github/workflows/v8upgrade.yml
```diff
@@ -0,0 +1,52 @@
...
+            - name: Fetch V8
+              run: cd deps && fetch v8 || true
+            - name: Get latest stable commit ref of v8
+              id: v8_latest_commit
+              run: echo ::set-output name=VALUE::$(curl -sS 'https://omahaproxy.appspot.com/all.json?os=linux&channel=stable' | jq '.[0]["versions"][0]["v8_commit"]')
+            - name: Get latest stable version number of v8
+              id: v8_latest_version
+              run: echo ::set-output name=VALUE::$(curl -sS 'https://omahaproxy.appspot.com/all.json?os=linux&channel=stable' | jq '.[0]["versions"][0]["v8_version"]')
+            - name: Checkout v8 to stable commit ref
+              run: cd deps/v8 && git fetch && git checkout ${{steps.v8_latest_commit.outputs.VALUE}}
```
dylanahsmith, line 34: Looks like we could just checkout the version number tag, to avoid needing to also get the commit SHA.  As in, it looks like this would work

```suggestion
              run: cd deps/v8 && git fetch && git checkout ${{steps.v8_latest_version.outputs.VALUE}}
```

genevieve: Is there actually a branch/tag for the full version? I thought that it was either checking out the v8 commit specifically or parsing the full version for the major+minor to compose the specific branch name: #8 in https://github.com/rogchap/v8go#upgrading-the-v8-binaries

deps/check_latest_v8_version.py
```diff
@@ -0,0 +1,31 @@
...
+
+# Script to compare current v8 version installed and the latest stable v8 version
+# Returns 1 if the versions are the same
+# Returns 0 if the latest stable version is higher than the current installed one
+
+import requests
+import sys
+
+# api-endpoint
+URL = "https://omahaproxy.appspot.com/all.json?os=linux&channel=stable"
```
dylanahsmith, line 11: Perhaps we should fetch all OS versions, filter the OSes that we are building for in python (linux, win64 and mac) and only upgrade only automatically upgrade when all the versions are the same.  If they have held off on promoting the version to stable for one of the OSes because of some problem, then this might help us to avoid it.

However, this might be fine for now.

deps/update_include_folder.py
```diff
@@ -0,0 +1,95 @@
...
+
+v8_include_directories_names = get_directories_names(v8_include_path)
+deps_include_directories_names = get_directories_names(deps_include_path)
+
+if not v8_include_directories_names == deps_include_directories_names:
+  new_directories = set(v8_include_directories_names).difference(set(deps_include_directories_names))
+  create_new_directories_and_vendor_files(new_directories, deps_include_path)
+
+delete_files_except_vendor_files(deps_include_path)
+copy_include_directory(v8_include_path, deps_include_path)
```
dylanahsmith, line 95: It might be simpler to just delete all the files in the destination include folder, since this script can already recreate the vendor.go files automatically.

```suggestion
shutil.rmtree(deps_include_path)
shutil.copytree(v8_include_path, deps_include_path, dirs_exist_ok=True)
create_new_vendor_files(get_directories_names(deps_include_path))
```

which would also simplify the above code

deps/check_latest_v8_version.py
dylanahsmith: Instead, I think this could be merged update_include_folder.py the update_v8_src.py script and could take on the work from several github action steps, making it easy to skip unnecessary following work and just not change any files.  The [peter-evans/create-pull-request](https://github.com/peter-evans/create-pull-request) github action documents that

> If there are no changes (i.e. no diff exists with the checked-out base branch), no pull request will be created and the action exits silently.

so we wouldn't even have to worry about it creating an empty PR.

.github/workflows/v8upgrade.yml
dylanahsmith: Yes, there is a tag.  For instance, this PRs deps/v8_version is 9.0.257.18 and there is a tag (https://chromium.googlesource.com/v8/v8.git/+/refs/tags/9.0.257.18).  If you `cd deps/v8 && git checkout 9.0.257.18` then it checks out that commit.

> Find the right branch-heads/** to checkout, for example if the v8_version is 8.7.220.31 then you want to git checkout branch-heads/8.7. You can check all the branch-heads with git branch --remotes | grep branch-heads/

It seems like the upgrade instructions would have led to not using the stable release version that it recommends looking up by checking out a branch instead.

rogchap: FYI: the branch naming upgrade instructions came from the official docs: https://v8.dev/docs/release-process#stable-releases

deps/check_latest_v8_version.py
rogchap: I've never seen this happen in practice, especially for mac and linux.

.github/workflows/v8upgrade.yml
```diff
@@ -0,0 +1,52 @@
+name: V8 Upgrade
+
+on:
+  schedule:
+    - cron: '0 0 * * *' # Run every day
```
rogchap, line 5: Stable releases are on average every 4 weeks. Running every day seems excessive, but I guess most of the time it will just be a version check.

.github/workflows/v8upgrade.yml
dylanahsmith: Oh, I guess they don't recommend the version chrome uses, since they aren't going to release fixes as frequently for a client application where the end user isn't going to want to upgrade their application that frequently.  So using the branch head might reduce the time to get security patches ready and released.

.github/workflows/v8upgrade.yml
dylanahsmith: > Roughly every 4 weeks a new **major** Stable release is done.

(Source: https://v8.dev/docs/release-process#stable-releases)

Emphasis on "major" was mine, they more frequently release patch version bumps.

Currently, https://omahaproxy.appspot.com shows the most recent version bump was only about a week apart:
* current_reldate: 09/30/21
* previous_reldate: 09/24/21

If we tracked the branch head for the stable channel (currently branch-heads/9.4) then that would make the changes even more frequent.  Releasing fixes that frequently might be overkill, but automating the preparation work for getting a security fix out would be useful.

GustavoCaso:
@rogchap @dylanahsmith @genevieve I've done some explorations on my fork and this is the conclusion I was able to gether.

- The new workflow `v8upgrade.yml` works as expected 
- The branch created by the workflow won't trigger any flows neither the `build` not the `test` ones. This is not an issue but rather a design from the GitHub actions [docs](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token)

> If you would like to trigger a workflow from a workflow run, you can trigger the event using a personal access token. You'll need to create a personal access token and store it as a secret. To minimize your GitHub Actions usage costs, ensure that you don't create recursive or unintended workflow runs. For more information on storing a personal access token as a secret, see "Creating and storing encrypted secrets."

- I tested on my fork by creating a PAT and it worked, but with the problem that the `test` workflow would also be triggered. This is unfortunate because they would always fail and it is a waste of time.
Possible solution: Would be to create a new base branch maybe `v8_upgrades` and make sure that the test workflow is not triggered on pull request or push which the target is that branch. 


Since these changes I believe are out of the scope of this PR, I think we can merge this one to start getting the benefits of the `upgrade` workflow and in the future, we can try improving it. 
Also, once this hits master I'm happy to continue working on improving the workflow.

Issue #105, opened 2021-04-13
Calling Go function callback from JS seems to leak memory


arctica:
Using the print function from the examples (even with the actual fmt.Printf() commented out) seems to result in a memory leak.

Slightly modified example from the Readme to call the print function in an infinite loop:

```Go
iso, _ := v8go.NewIsolate() // create a new VM
// a template that represents a JS function
printfn, _ := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *v8go.Value {
    //fmt.Printf("%v", info.Args()) // when the JS function is called this Go callback will execute
    return nil // you can return a value back to the JS caller if required
})
global, _ := v8go.NewObjectTemplate(iso) // a template that represents a JS Object
global.Set("print", printfn) // sets the "print" property of the Object to our function
ctx, _ := v8go.NewContext(iso, global) // new Context with the global Object set to our object template
ctx.RunScript("while (true) { print('foo') }", "print.js") // will execute the Go callback with a single argunent 'foo'
```

**Warning: Running this will quickly consume all memory of the machine.**

Removing the 'foot' parameter of the print() call seems to stop the leak. So I guess the parameters of the callback are somehow leaking.

arctica:
I think v8go.NewValue() also leaks memory, even if you don't pass it back as a result from the Go callback.

```Go
iso, _ := v8go.NewIsolate()
fn, _ := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *v8go.Value {
    v8go.NewValue(iso, "foo")
    return nil
})

global, _ := v8go.NewObjectTemplate(iso)
global.Set("fn", fn)
ctx, _ := v8go.NewContext(iso, global)
ctx.RunScript("while (true) { fn(); }", "fn.js")
```

rogchap:
New Values are attached to the Isolate that created them, you need to Dispose the isolate to free any value's memory. For the Function the Values should be attached to the Context that created them, so you need to call `ctx.Close()` to free the memory.

Maybe we need an explicit call to free the values, if you want to free the memory without closing or disposing of the isolate or context? 🤔

arctica:
It's impractical to free their memory only when the isolate or context is closed. Not every VM is short lived. In my case I have a bunch of very long running isolates/contexts that are fed new values from Go as events arrive.

I see two ways to do this:

1. Copy the values somehow into v8 and let it manage those with its GC.
[...]

PR #182, opened 2021-09-27
Fix crashes from template finalizer

dylanahsmith:
## Problem

We ran into fairly reproducible crashes from benchmarking code that was creating (and initializing) new isolates frequently, often in C.TemplateSetTemplate, where the value being set was a v8go.FunctionTemplate.  At first, I thought this might have been from a race condition from the finalizer running in a separate goroutine, which I thought was the reason other uses of finalizers was removed from v8go, but it happened even when the finalizer only freed the wrapper struct (m_template).  It looks like there was also a problem with the `template` becoming garbage after `template.ptr` is obtained from it, even before `template.ptr` is passed into the C function (e.g. `C.TemplateSetTemplate`).

## Solution

To avoid the finalizer goroutine race condition, I changed the finalizer to only free the template wrapper struct.   v8::Template resides in the V8 heap (i.e. it is v8::Data), so it should get freed when the isolate is disposed.  In the future, the solution for #105 should allow this v8::Data to be released earlier.

The second commit in this PR uses runtime.KeepAlive to avoid the template Go struct from being garbage collected before the `template.ptr` is finished being used.

I tested this in our application benchmark that was crashing and this seems to have fixed the crashes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Add some missing error propagation
 - Fix crash from template finalizer releasing V8 data, let it be disposed with the isolate
+- Fix crash by keeping alive the template while its C++ pointer is still being used
 
 ## [v0.6.0] - 2021-05-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Add some missing error propagation
+- Fix crash from template finalizer releasing V8 data, let it be disposed with the isolate
 
 ## [v0.6.0] - 2021-05-11
 
--- a/template.go
+++ b/template.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"runtime"
 	"unsafe"
 )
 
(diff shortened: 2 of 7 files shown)


template.go
```diff
@@ -57,7 +59,9 @@ func (t *template) Set(name string, val interface{}, attributes ...PropertyAttri
 }
 
 func (t *template) finalizer() {
-	C.TemplateFree(t.ptr)
+	// Using v8::PersistentBase::Reset() wouldn't be thread-safe to do from
+	// this finalizer goroutine so just free the wrapper and let the template
+	// itself get cleaned up when the isolate is disposed.
+	C.TemplateFreeWrapper(t.ptr)
 	t.ptr = nil
-	runtime.SetFinalizer(t, nil)
```
dylanahsmith, line 62: From https://pkg.go.dev/runtime#SetFinalizer

> When the garbage collector finds an unreachable block with an associated finalizer, it clears the association and runs finalizer(obj) in a separate goroutine. This makes obj reachable again, but now without an associated finalizer.

so Go should will clear the finalizer for us, thus we don't need to do that here.

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Add some missing error propagation
+- Fix crash from template finalizer releasing V8 data, let it be disposed with the isolate
+- Fix crash keeping alive the template while its C++ pointer is still being used
```
genevieve, line 28: "Fix crash _by_ keeping alive" ?

genevieve (approved):

Issue #186, opened 2021-09-28
Function callbacks panic inside of `JSONStringify`


colinking:
We're running into a crash with v8go when using `.JSONStringify()`. I put together a minimal example ([full code here](https://github.com/colinking/v8-experiments/blob/main/proxy-crash/main.go)):

<details>

```go
package main

import (
	"fmt"

	// This example uses our fork for darwin_arm64 support, but the issue
	// should surface using the upstream master, too:
	// v8 "rogchap.com/v8go"
	v8 "github.com/airplanedev/v8go"
)

var code = `
	new Proxy({}, {
		get: function () {
			foobar()
			return "hello!"
		},
	})
`

func main() {
	iso := v8.NewIsolate()
	defer iso.Dispose()

	global := v8.NewObjectTemplate(iso)
	err := global.Set("foobar", v8.NewFunctionTemplate(iso, func(info *v8.FunctionCallbackInfo) *v8.Value {
		v8v, _ := v8.NewValue(iso, "callback")
		return v8v
	}))
	if err != nil {
		panic(fmt.Errorf("setting global: %w", err))
	}

	v8ctx := v8.NewContext(iso, global)
	defer v8ctx.Close()

	v8v, err := v8ctx.RunScript(code, "main.js")
	if err != nil {
		panic(fmt.Errorf("running script: %w", err))
	}

	s, err := v8.JSONStringify(nil, v8v)
	if err != nil {
		panic(fmt.Errorf("stringifying value: %w", err))
	}
	fmt.Println(s)
}
```

</details>

If you run it with `go run ./main.go`, it'll produce the following crash log:

<details>

```
❯ go run ./main.go

#
# Fatal error in v8::Isolate::Dispose()
# Disposing the isolate that is entered by a thread.
#

SIGTRAP: trace trap
PC=0x104a81fe8 m=0 sigcode=0

goroutine 0 [idle]:
runtime: unknown pc 0x104a81fe8
stack: frame={sp:0x16b4333d0, fp:0x0} stack=[0x16b3b43e8,0x16b433460)
0x000000016b4332d0:  0x0001000200030003  0x0000000000000000
0x000000016b4332e0:  0x0000000000000000  0x0000000000000000
0x000000016b4332f0:  0x0000000000000000  0x0000000000000000
0x000000016b433300:  0x0000000000000000  0x0000000000000000
0x000000016b433310:  0x0000000000000000  0x0000000000000000
[...]

PR #187, opened 2021-09-29
Fix function callback usage outside of `RunScript`

colinking:
This fixes an issue where you could not access function callbacks outside of `RunScript` since the context registry would not have been updated.

Fixes: https://github.com/rogchap/v8go/issues/186

### DIFF
--- a/context.go
+++ b/context.go
@@ -16,8 +16,6 @@ import (
 // Due to the limitations of passing pointers to C from Go we need to create
 // a registry so that we can lookup the Context from any given callback from V8.
 // This is similar to what is described here: https://github.com/golang/go/wiki/cgo#function-variables
-// To make sure we can still GC *Context we register the context only when we are
-// running a script inside the context and then deregister.
 type ctxRef struct {
 	ctx      *Context
 	refCount int
@@ -73,6 +71,7 @@ func NewContext(opt ...ContextOption) *Context {
 		ptr: C.NewContext(opts.iso.ptr, opts.gTmpl.ptr, C.int(ref)),
 		iso: opts.iso,
 	}
+	ctx.register()
 	runtime.KeepAlive(opts.gTmpl)
 	return ctx
 }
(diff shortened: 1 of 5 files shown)


colinking:
Just updated this PR with a test case for the behavior in #186

cc @dylanahsmith and @rogchap since [you two likely have the most context here](https://github.com/rogchap/v8go/commits/master/context.go)

context_test.go
```diff
@@ -118,6 +113,51 @@ func TestMemoryLeak(t *testing.T) {
...
+
+	iso := v8.NewIsolate()
+	defer iso.Dispose()
+
+	global := v8.NewObjectTemplate(iso)
+	err := global.Set("location", v8.NewFunctionTemplate(iso, func(info *v8.FunctionCallbackInfo) *v8.Value {
+		v, err := v8.NewValue(iso, "world")
+		if err != nil {
+			t.Errorf("creating return value: %v", err)
+		}
```
dylanahsmith, line 128: Other than the last assertion, all the others would just result in following test errors, so they may as well be fatal errors for the test.

There is a failIf helper that can be used instead to simplify these

```suggestion
		failIf(t, err)
```

dylanahsmith (approved):
The other benefit of registering and deregistering each context once is that it reduces lock content, so was already something I was planning on proposing.

colinking:
Thanks for the speedy review @dylanahsmith. I just updated the test's error handling, if you can take another look.

context.go
```diff
@@ -16,8 +16,6 @@ import (
 // Due to the limitations of passing pointers to C from Go we need to create
 // a registry so that we can lookup the Context from any given callback from V8.
 // This is similar to what is described here: https://github.com/golang/go/wiki/cgo#function-variables
-// To make sure we can still GC *Context we register the context only when we are
```
dylanahsmith, line 19: Ah, this is why it didn't do this before, which doesn't apply now that there is no GC finalizer for Context.  I guess we might need to surround v8 calls in the future again if we decide to bring back Context GC

dylanahsmith:
Thanks for fixing this!  Otherwise, I'm sure we would run into the same problem

PR #190, opened 2021-09-30
Mention context registry fix in changelog

colinking:
This is a follow-up to https://github.com/rogchap/v8go/pull/187 where I had forgotten to add an entry to the changelog.

cc @dylanahsmith


dylanahsmith (approved):
Thanks, I need to get better and remembering to check for changelog entries

PR #195, opened 2021-10-13
Support FunctionCallback returning an error, and native Exceptions.

tommie:
**DRAFT: Awaiting #274 to refactor `JSError`.**

Returning an error from a callback feels like the most Go way of supporting exceptions, and it ensures the Go code doesn't continue executing after throwing (as it would if we instead supported `Isolate.ThrowException`).

This also adds Error constructors, so functions can throw real JS Errors. Note that the new callback interface does allow throwing arbitrary objects, but `*Exception` is the only built-in object that is both a `Valuer` and `error`.

### DIFF
--- a/v8go.cc
+++ b/v8go.cc
@@ -910,7 +910,9 @@ RtnValue NewValueBigIntFromWords(IsolatePtr iso,
   return rtn;
 }
 
-ValuePtr NewValueError(IsolatePtr iso, ErrorTypeIndex idx, const char* message) {
+ValuePtr NewValueError(IsolatePtr iso,
+                       ErrorTypeIndex idx,
+                       const char* message) {
   ISOLATE_SCOPE_INTERNAL_CONTEXT(iso);
   Local<Context> local_ctx = ctx->ptr.Get(iso);
   Context::Scope context_scope(local_ctx);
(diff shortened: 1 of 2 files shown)


exception.go
```diff
@@ -0,0 +1,109 @@
...
+	// #include <stdlib.h>
+	// #include "v8go.h"
+	"C"
+
+	"fmt"
+	"unsafe"
+)
+
+// NewRangeError creates a RangeError.
+func NewRangeError(iso *Isolate, msg string) *Exception {
```
dylanahsmith, line 17: There is already the JSError type which is used to return exceptions from v8go functions.  It seems like there should be consistency between the exception value returned by other v8go functions and these functions to construct errors.  That should make it easier to propagate exceptions, even without adding a new feature to return errors from function callbacks.

Currently, JSError is converting the underlying JS error object to a string and storing it in the Message field.  Instead, it should store the underlying JS error object, so that the error can be propagated.  That Message field could be deprecated in favour of a method so that the conversion can be done on-demand in the future.  It looks like the stack trace can similarly be obtained on-demand using [v8::Exception::GetStackTrace](https://v8.github.io/api/head/classv8_1_1Exception.html#a47e3613ae924508cc0ca29ad05e2464a).

The JSError.Location field is derived from the a v8::Message on v8::TryCatch, which may be present without a stack trace for compilation errors.  v8::Exception::CreateMessage can be used to obtained a v8::Message from an exception value, but we don't want to automatically create it in a getter method.  Since it is a public field, it will need to continue populating on creation for the current code paths, but it could similarly be deprecated in favour of a method to obtain this on-demand for new code paths.  This on-demand code path could at least obtain the when the error has a stack trace, even if we don't get it to initially work on this new code path errors with a location but no stack trace.

It seems like this refactoring and the addition of these functions to create exceptions could be split out into a separate PR.

dylanahsmith: I've opened this as an issue (https://github.com/rogchap/v8go/issues/274) for greater visibility on this issue with JSError

tommie: I don't remember exactly why I created a new type for it, but it was probably to avoid having to refactor `JSError` (avoiding merge conflicts). If there's consensus that's better, that sounds good to me. Thanks for reviewing.

PR #206, opened 2021-10-26
Add functions to separate compilation from running to speed up start times on new isolates loading the same scripts.

genevieve:
### Problem
The original problem we were trying to solve was amortizing the cost of running some of the same scripts needed to prep isolates/contexts in a pool that we manage. After starting down this effort, we noticed that `RunScript` on contexts that belong to different isolates seem to be run much faster as a result of (likely) a shared compiler cache by v8 across different isolates. We can see this when we run the commands subsequently in completely different isolates and see the run time drastically improve. For instance, in the performance block below, we saw that the first `RunScript` in isolate #1 takes 369 us and the subsequent call in isolate #2 (using the same script) runs in 45 us. Similarly for our compiled/run split code.
However, this only works when the isolates are running in the same process (either synchronously or async via goroutines) so in a scenario where there are different processes for new isolates (i.e. for improved security in the form of sandboxing), there's no sharing of the compiler cache so it would be beneficial to split the compilation step from the run step (which is what `RunScript` is doing.) Additionally, when isolates are launched in parallel in goroutines, the time it takes to run the same scripts in their contexts almost doubles as a result of what appears to be v8 synchronizing access to the shared compiler cache. So while these isolates might be launched in the same process, the start time to use them (with the scripts loaded) is nearly double so this scenario would similarly benefit from a separation of compilation of a script and running it.

### Solution

This PR adds a function to `Isolate` called `CompileUnboundScript` which takes the source code, origin/filename, and a set of compile options to produce an unbound script. The compile options support either providing a reference to the new `ScriptCompilerCachedData` type and/or a compile option that maps to `v8::ScriptCompiler::CompileOptions` enum. On the `UnboundScript`, there are two functions introduced called `Run(ctx)` (which binds the script to that context and runs it so long as that context belongs to the same isolate the script was compiled in) and `CreateCodeCache()` which returns the new `ScriptCompilerCachedData` type which just contains the `Bytes` of the data and the `Rejected` flag indicating whether it was used or not during compilation. This cached data can be used to compile new unbound scripts in different isolates.

~~This PR adds a function to `Isolate` called `CompileScript` which takes the source code and the origin to compile the script and return `[]byte` of the compiled script. This value can then be passed to a new function on `Context` called `RunCompiledScript`. In the first, an unbound script is created. In the second, the unbound script is bound to the provided context and executed.~~

~~We considered a few API variations~~
- ~~Exposing `Script` and `UnboundScript` directly so users would generate an UnboundScript, then call `BindToCurrentContext` which returns a `Script` which can then be used in `Run` (which would actually run that script.) This had challenges and downsides - UnboundScript still belongs to an isolate so the isolate used to generate the compiled data that can be used in new isolates/contexts would need to be persisted and we could not discard that original isolate. Similarly, Script is bound to a context so on every new context, the unbound script would have to be bind with a v8go api call before being run. Since its required to bind an unbound script to a context before running it, it seems unnecessary to expose this additional api call to users. Without that call, there's no reason to introduce the Script type to v8go. Second, in the case that isolates are launched in their own processes, the Script pointer can't be passed to them for launch in the same way that a byte slice could on stdin to be read and passed back into a v8go call (like RunCompileScript accepts.)~~
- ~~Exposing `ScriptCompiler::CachedData*` directly. This would have the benefit of not copying the []byte slice on compilation but in order to get this cached data into a separate process to run it in a new isolate, it would need to be loaded anyway so it could be passed to the process.~~

### Questions/Thoughts
- ~~CachedData has a `rejected` field that indicates it wasn't used (for any number of reasons) when part of the Source that generates an UnboundScript before it becomes a Script and is Run bound to a context. It might be useful to expose this as an error?~~
- Benchmarks - since isolates in the same process are sharing the v8 compiler's cached data, it was hard to ensure that we were in fact using the compiled cached data to `RunCompiledScript` and not the implicit cache. There's a toy project [here](https://github.com/genevieve/v8cachedwut/blob/main/result) that passes the cached data from one process as a byte slice on stdin to another process and measures the performance time. There's an additional [branch](https://github.com/genevieve/v8cachedwut/blob/gl-sep/result) on the repo that shows that `RunScript` takes the same time for the same script to run in separate processes (just to emphasize the time difference when compilation + run happens together.) We could add a benchmark to this function but the value might not be totally apparent.

#### Performance
```
# Measuring `ctx.RunScript`
time to RunScript in i1: 369.972µs
time to RunScript in i2: 45.958µs

# Measuring `iso.CompileScript`
go test -run TestIsolateCompileScript
time to compile script in i1: 318.222µs
time to compile script in i2: 143.299µs
time to compile script in i3: 162.701µs

# Run compiled script in isolates
# NOTE: for our tiny script, there seems to be a fastest speed of running around 10 us
go test -run TestIsolateCompileScript
time to compile script in i1: 569.07µs
time to run compiled script in i2: 123.079µs
time to run compiled script in i3: 26.947µs
time to run compiled script in i4: 17.99µs
time to run compiled script in i5: 13.542µs

# Changing order of isolates running scripts to see if it matters
# NOTE there seems to be something weird around ordering isolates
go test -run TestIsolateCompileScript
time to compile script in i1: 359.817µs
time to run compiled script in i1: 74.715µs
time to run compiled script in i2: 16.568µs
time to run compiled script in i3: 11.392µs

# Comparing heap statistics
# Note that the TotalPhysicalSize is greater in the isolate that runs the compiled script 
# than the one that compiles it (likely due to the actual execution of the script contents) 
# but PeakMallocedMemory is far greater in the isolate that compiled the script than ran it
i1: {TotalHeapSize:1323008 TotalHeapSizeExecutable:262144 TotalPhysicalSize:444572 TotalAvailableSize:1518012192 UsedHeapSize:742964 HeapSizeLimit:1518338048 MallocedMemory:16476 ExternalMemory:0 PeakMallocedMemory:8192 NumberOfNativeContexts:2 NumberOfDetachedContexts:0}
i2: {TotalHeapSize:1323008 TotalHeapSizeExecutable:262144 TotalPhysicalSize:444572 TotalAvailableSize:1518012192 UsedHeapSize:742964 HeapSizeLimit:1518338048 MallocedMemory:16476 ExternalMemory:0 PeakMallocedMemory:8192 NumberOfNativeContexts:2 NumberOfDetachedContexts:0}
time to compile script in i1: 374.224µs
time to run compiled script in i2: 76.937µs
i1: {TotalHeapSize:1323008 TotalHeapSizeExecutable:262144 TotalPhysicalSize:446660 TotalAvailableSize:1518010536 UsedHeapSize:744620 HeapSizeLimit:1518338048 MallocedMemory:16476 ExternalMemory:0 PeakMallocedMemory:24576 NumberOfNativeContexts:2 NumberOfDetachedContexts:0}
i2: {TotalHeapSize:1323008 TotalHeapSizeExecutable:262144 TotalPhysicalSize:446860 TotalAvailableSize:1518010536 UsedHeapSize:744620 HeapSizeLimit:1518338048 MallocedMemory:16476 ExternalMemory:0 PeakMallocedMemory:8192 NumberOfNativeContexts:2 NumberOfDetachedContexts:0}
```


wisepythagoras (approved):

genevieve:
@rogchap I'd be interested in fleshing out this proposed API change a bit more as neither the existing RunScript or the new RunCompiledScript/CompileScript map to v8 apis.

To be an "exact" match, we'd want to expose:
- v8::ScriptCompiler::CompileUnboundScript
- v8::ScriptCompiler::Script (which gets returned from ^ and internally has it's code cache pointer)
- v8::Script::BindToCurrentContext
- v8::Script::Run (we could skip the bind function and require a context be passed as a argument to the v8go function we expose)

thus a user would do:
```go
script := iso.CompileUnboundScript(script, origin)
err := script.BindToCurrentContext(ctx)
val, err := script.Run()

# or

val, err := ctx.RunScript(source, origin) 
```

If we're just trying to achieve the more natural compilation / run split that v8 has, we can rename the functions:

- `ctx.RunScript(source, origin)` -> `ctx.CompileAndRun(source, origin)`
- `iso.CompileScript(source, origin)` stays the same and continues to return `cachedData`
- `ctx.RunCompiledScript(source, cachedData, origin)` -> `ctx.RunScript(source, cachedData, origin)`

rogchap:
I think the latter makes more sense. We want to align with the functionality of V8 as closely as possible, but we don't need to be exact. 
By doing the later we still give the caller control over the execution and it's still intuitive.

README.md
```diff
@@ -21,9 +21,9 @@ import v8 "rogchap.com/v8go"
 
 ```go
 ctx := v8.NewContext() // creates a new V8 context with a new Isolate aka VM
-ctx.RunScript("const add = (a, b) => a + b", "math.js") // executes a script on the global context
-ctx.RunScript("const result = add(3, 4)", "main.js") // any functions previously added to the context can be called
-val, _ := ctx.RunScript("result", "value.js") // return a value in JavaScript back to Go
+ctx.CompileAndRun("const add = (a, b) => a + b", "math.js") // executes a script on the global context
+ctx.CompileAndRun("const result = add(3, 4)", "main.js") // any functions previously added to the context can be called
+val, _ := ctx.CompileAndRun("result", "value.js") // return a value in JavaScript back to Go
```
genevieve, line 26: @rogchap an alternative here could be `RunUncompiledScript` to differ more clearly from `RunScript`

script_compiler.go
```diff
@@ -0,0 +1,15 @@
...
+// found in the LICENSE file.
+
+package v8go
+
+type ScriptCompilerCachedData []byte
+
+type ScriptCompilerCompileOption int
+
+const (
+	ScriptCompilerCompileOptionNoCompileOptions = iota
```
dylanahsmith, line 12: This seems quite verbose for something that would ideally just be a default argument that wouldn't need to be specified.

genevieve: This was addressing a comment you made here: https://github.com/Shopify/v8go/pull/13#discussion_r729825955
I agree that it's the default but it is different than `EagerCompile` and I wanted to have our api be consistent with v8 in this area. Specifically, the integer mapping from this enum is 1:1 with the enum in v8.

genevieve: The `ScriptCompiler` prefix is verbose but it would align again with the suggestion you made to call `CachedData` -> `ScriptCompilerCachedData`

dylanahsmith: It looks like this gets referred to as `v8::ScriptCompiler::kNoCompileOptions` when using the V8 API, which avoids the redundant `CompileOption` reference in the name. `ScriptCompilerNoCompileOptions` would actually be closer to how the API gets used in C++.

I do think `CachedData` could be ambiguous.  `ScriptCompilerCachedData` does seem verbose, although at least doesn't introduce redundancy.  Naming it `CompilerCachedData` seems like a better name with the current C++ API, given that function and module compilation also happens on ScriptCompiler.  I'm just not sure why the C++ API didn't name `ScriptCompiler` as `Compiler`.

isolate.go
```diff
@@ -75,6 +77,53 @@ func (i *Isolate) IsExecutionTerminating() bool {
...
+	cOptions := C.CompileOptions{
+		compileOption: C.int(opts.Option),
+	}
+	if opts.CachedData != nil {
+		cOptions.cachedData = C.ScriptCompilerCachedData{
+			data:   (*C.uchar)(unsafe.Pointer(&opts.CachedData.Bytes[0])),
+			length: C.int(len(opts.CachedData.Bytes)),
+		}
+		if opts.CachedData.Rejected {
+			cOptions.cachedData.rejected = C.int(1)
```
genevieve, line 110: `C.IsolateCompileUnboundScript` doesn't actually use the incoming `cOptions.cachedData.rejected` field because whether it was rejected before or not by some other usage is irrelevant to whether it will be useful to this call so this can probably be removed.

isolate.go
```diff
@@ -75,6 +77,53 @@ func (i *Isolate) IsExecutionTerminating() bool {
...
+
+	cOptions := C.CompileOptions{
+		compileOption: C.int(opts.Option),
+	}
+	if opts.CachedData != nil {
+		cOptions.cachedData = C.ScriptCompilerCachedData{
+			data:   (*C.uchar)(unsafe.Pointer(&opts.CachedData.Bytes[0])),
+			length: C.int(len(opts.CachedData.Bytes)),
+		}
+		if opts.CachedData.Rejected {
```
genevieve, line 109: C.IsolateCompileUnboundScript doesn't actually use the incoming cOptions.cachedData.rejected field because whether it was rejected before or not by some other usage is irrelevant to whether it will be useful to this call so this can probably be removed.

dylanahsmith (approved):

PR #221, opened 2021-10-31
Add the ThrowException fn to the isolate

wisepythagoras:
I've re-implemented the `ThrowException` function from #143, but added support for passing in a `Value` instead of just a string, as pointed out by @dylanahsmith. This allows us to pass a string, or forward in a `TypeError`.

A note: The return value is always `null` or `undefined`.

### DIFF
--- a/isolate.go
+++ b/isolate.go
@@ -103,9 +103,10 @@ func (i *Isolate) Dispose() {
 	i.ptr = nil
 }
 
-// ThrowException will schedule for an exception to be thrown. This prevents
-// any JavaScript to be invoked and the caller must run immediately and the
-// JavaScript will only run after the exeption is handled.
+// ThrowException schedules an exception to be thrown when returning to
+// JavaScript. When an exception has been scheduled it is illegal to invoke
+// any JavaScript operation; the caller must return immediately and only after
+// the exception has been handled does it become legal to invoke JavaScript operations.
 func (i *Isolate) ThrowException(value *Value) *Value {
 	return &Value{
 		ptr: C.ThrowException(value.ptr),
--- a/isolate.go
+++ b/isolate.go
@@ -103,6 +103,13 @@ func (i *Isolate) Dispose() {
 	i.ptr = nil
 }
 
+// ThrowException will schedule for an exception to be thrown.
+func (i *Isolate) ThrowException(value *Value) *Value {
+	return &Value{
+		ptr: C.ThrowException(value.ptr),
+	}
+}
+
 // Deprecated: use `iso.Dispose()`.
 func (i *Isolate) Close() {
 	i.Dispose()
--- a/isolate.go
+++ b/isolate.go
@@ -109,7 +109,7 @@ func (i *Isolate) Dispose() {
 // the exception has been handled does it become legal to invoke JavaScript operations.
 func (i *Isolate) ThrowException(value *Value) *Value {
 	return &Value{
-		ptr: C.ThrowException(value.ptr),
+		ptr: C.IsolateThrowException(i.ptr, value.ptr),
 	}
 }
 
--- a/isolate.go
+++ b/isolate.go
@@ -103,7 +103,9 @@ func (i *Isolate) Dispose() {
 	i.ptr = nil
 }
 
-// ThrowException will schedule for an exception to be thrown.
+// ThrowException will schedule for an exception to be thrown. This prevents
+// any JavaScript to be invoked and the caller must run immediately and the
+// JavaScript will only run after the exeption is handled.
 func (i *Isolate) ThrowException(value *Value) *Value {
 	return &Value{
 		ptr: C.ThrowException(value.ptr),
--- a/isolate.go
+++ b/isolate.go
@@ -108,6 +108,9 @@ func (i *Isolate) Dispose() {
 // any JavaScript operation; the caller must return immediately and only after
 // the exception has been handled does it become legal to invoke JavaScript operations.
 func (i *Isolate) ThrowException(value *Value) *Value {
+	if i.ptr == nil {
+		panic("Isolate has been disposed")
+	}
 	return &Value{
 		ptr: C.IsolateThrowException(i.ptr, value.ptr),
 	}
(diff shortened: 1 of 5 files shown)


genevieve:
Might be able to test it by calling the function from within another function like:
```go
func TestIsolateThrowException(t *testing.T) {
	t.Parallel()
	iso := v8.NewIsolate()
	defer iso.Dispose()

	typeErr, err := v8.NewValue(iso, "some type error")
	if err != nil {
		t.Errorf("unexpected error: %v", err)
	}

	fn := v8.NewFunctionTemplate(iso, func(info *v8.FunctionCallbackInfo) *v8.Value {
		v := iso.ThrowException(typeErr)
		if !v.IsNull() && !v.IsUndefined() {
			t.Error("expected result to be <nil> or undefined")
		}
		return nil
	})

	global := v8.NewObjectTemplate(iso)
	global.Set("foo", fn)

	ctx := v8.NewContext(iso, global)
	defer ctx.Close()

	_, e := ctx.RunScript("foo()", "forever.js")
	if e.Error() != "some type error" {
		t.Errorf("expected error but got: %v", e)
	}
}
```

wisepythagoras:
Thanks, @genevieve. I had forgotten about the error returned by `RunScript`. I modified it a little to include a test case for a `TypeError`.

isolate.go
```diff
@@ -103,6 +103,13 @@ func (i *Isolate) Dispose() {
 	i.ptr = nil
 }
 
+// ThrowException will schedule for an exception to be thrown.
```
genevieve, line 106: Could you update the Changelog to reflect this new function?

genevieve: Additionally, I wonder how much of the v8 comment for this function should be captured here?
```
   * Schedules an exception to be thrown when returning to JavaScript.  When an
   * exception has been scheduled it is illegal to invoke any JavaScript
   * operation; the caller must return immediately and only after the exception
   * has been handled does it become legal to invoke JavaScript operations.
```

isolate_test.go
```diff
@@ -135,6 +135,68 @@ func TestIsolateGarbageCollection(t *testing.T) {
...
+	ctx := v8.NewContext(iso, global)
+	defer ctx.Close()
+
+	_, e := ctx.RunScript("foo()", "forever.js")
+
+	if e.Error() != "some type error" {
+		t.Errorf("expected \"some type error\" error but got: %v", e)
+	}
+
+	_, e = ctx.RunScript("foo2(new TypeError('this is a test'))", "forever.js")
```
genevieve, line 193: The filenames for these scripts seem to have been copied from another test where the source code was actually a forever-looping function. We can rename this to just `foo.js`

isolate_test.go
```diff
@@ -135,6 +135,68 @@ func TestIsolateGarbageCollection(t *testing.T) {
...
 }
 
+func TestIsolateThrowException(t *testing.T) {
+	t.Parallel()
+	iso := v8.NewIsolate()
+	defer iso.Dispose()
+
+	strErr, err := v8.NewValue(iso, "some type error")
+
+	if err != nil {
```
genevieve, line 145: We can assume that `v8.NewValue` won't return an error here (which is the pattern in the other tests because we can test the error handling in just `value_test.go` instead of everywhere)

isolate_test.go
```diff
@@ -135,6 +135,68 @@ func TestIsolateGarbageCollection(t *testing.T) {
...
+		return nil
+	})
+
+	// Function that is passed a TypeError from JavaScript.
+	fn2 := v8.NewFunctionTemplate(iso, func(info *v8.FunctionCallbackInfo) *v8.Value {
+		args := info.Args()
+
+		if len(args) < 1 {
+			t.Error("expected a TypeError argument")
+		}
```
genevieve, line 171: Since the unit test is the only one using this function, we can probably assume that an argument is passed in. If we were testing the function template and info args behavior, it would make sense to check these but we cover that in `function_template_test.go`

isolate.go
wisepythagoras: Yup! That makes sense. I've pushed a commit with all these fixes.

v8go.cc
```diff
@@ -198,6 +198,25 @@ IsolateHStatistics IsolationGetHeapStatistics(IsolatePtr iso) {
                             hs.number_of_detached_contexts()};
 }
 
+/********** Exceptions & Errors **********/
+
+ValuePtr ThrowException(ValuePtr value) {
+  Isolate* iso = value->iso;
+  m_ctx* ctx = value->ctx;
```
genevieve, line 205: One convention that we may want to follow is that the function in v8go.cc is usually named with the object it belongs to as a prefix and as it's first argument, like:
```cpp
void IsolateDispose(IsolatePtr iso) {
  if (iso == nullptr) {
    return;
  }
  ContextFree(isolateInternalContext(iso));
  iso->Dispose();
}
```
i.e.
```cpp
ValuePtr IsolateThrowException(IsolatePtr iso, ValuePtr value) {
  if (iso == nullptr) {
    return;
  }
...
}
```
then in the Go function:
```go
func (i *Isolate) ThrowException(value *Value) *Value {
	return &Value{
		ptr: C.ThrowException(i.ptr, value.ptr),
	}
}
```
This way we don't rely on the value's field to get an isolate and risk a nil pointer error because the isolate in go is a pointer that the function `ThrowException` would be called on and it would fail at the go level if the isolate is nil instead of within the c++ code. (There are probably other reasons that are slipping my mind right now.)

genevieve (approved):

v8go.cc
```diff
@@ -198,6 +198,28 @@ IsolateHStatistics IsolationGetHeapStatistics(IsolatePtr iso) {
                             hs.number_of_detached_contexts()};
 }
 
+/********** Exceptions & Errors **********/
+
+ValuePtr IsolateThrowException(IsolatePtr iso, ValuePtr value) {
+  if (iso == nullptr) {
+    return nullptr;
+  }
```
dylanahsmith: We should do this check in the calling Go code, since it should panic to call methods like this on a disposed isolate.  I've pushed a commit to do this.

dylanahsmith (approved):

v8go.cc
genevieve: Do you want to add a test for your fix using the helper recoverPanic function?

PR #226, opened 2021-11-09
Add Leak Check workflow to CI that uses Leak Sanitizer

### LINKED ISSUE #222: Add leak checking to CI
Instead of trying to [run valgrind](https://github.com/rogchap/v8go/issues/218#issuecomment-961302410) across different platforms, we can conditionally set flags for relevant clang sanitizers on the builds of v8go when we are running tests so that we can automate the checking of any memory/thread leakage (for example.)

We could do the conditional configuration in here: https://github.com/rogchap/v8go/blob/master/cgo.go

Here is an example clang sanitizer: https://clang.llvm.org/docs/LeakSanitizer.html

dylanahsmith:
Fixes #222

I added a separate workflow that does leaking, since it seems redundant to do on multiple platforms or go versions.  It also seemed simpler than adding conditionals to fit it into the existing test workflow.

I've tried using this locally on Linux and MacOS and have added some README documentation for doing leak checking locally.  I haven't tested on windows, but it looks like it currently isn't supported on windows (https://clang.llvm.org/docs/LeakSanitizer.html#supported-platforms)

### DIFF
--- a/.github/workflows/leakcheck.yml
+++ b/.github/workflows/leakcheck.yml
@@ -1,0 +1,22 @@
+name: Leak Check
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17.2
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Go Test
+      run: go test -v --tags leakcheck ./...
--- a/.github/workflows/leakcheck.yml
+++ b/.github/workflows/leakcheck.yml
@@ -19,4 +19,9 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
     - name: Go Test
-      run: go test -v --tags leakcheck ./...
+      env:
+        CC: clang
+        CXX: clang++
+      run: |
+        go test -c --tags leakcheck
+        ./v8go.test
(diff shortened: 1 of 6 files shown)


genevieve (approved):

dylanahsmith:
I pushed another commit now that I actually got line numbers in the backtrace for a reported leak, on both macOS and ubuntu.

These are similar workarounds to what I had to do for debugging crashes, which is already documented in the README:

> Build the executable to debug, using go build for commands or go test -c for tests. You may need to add the -ldflags=-compressdwarf=false option to disable debug information compression so this information can be read by the debugger (e.g. lldb that comes with Xcode v12.5.1, the latest Xcode released at the time of writing)

Disabling dwaft compression only seemed to be needed on macOS, but separating the compile and run of the test executable was needed on both.  I also had to use clang on CI, since even though gcc was able to catch the leak, it provided even less debugging information.

genevieve:
Wanted to confirm that the mac os instructions worked for me on a commit that contained a leak:

```
➜  v8go git:(b0165e5) CXX=/usr/local/opt/llvm/bin/clang++ CC=/usr/local/opt/llvm/bin/clang go test -c --tags leakcheck -ldflags=-compressdwarf=false
➜  v8go git:(b0165e5) ASAN_OPTIONS=detect_leaks=1 ./v8go.test
v8go.test(29413,0xfae9600) malloc: nano zone abandoned due to inability to preallocate reserved vm space.
some string
PASS

=================================================================
==29413==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 24 byte(s) in 1 object(s) allocated from:
    #0 0x57fed9d in wrap__Znwm+0x7d (libclang_rt.asan_osx_dynamic.dylib:x86_64+0x50d9d)
    #1 0x416e324 in UnboundScriptCreateCodeCache v8go.cc:636
    #2 0x41613b7 in _cgo_5cf3a7ef99fb_Cfunc_UnboundScriptCreateCodeCache cgo-gcc-prolog:65
    #3 0x40709ef in runtime.asmcgocall asm_amd64.s:667

SUMMARY: AddressSanitizer: 24 byte(s) leaked in 1 allocation(s).
```

PR #230, opened 2021-11-15
Free memory on long lived isolates and contexts

jblernout:
It is beneficial to create an isolate and a context only once and use them several times instead of recreating them each time. 

Creating an isolate and a context can take 3ms which is significant in some contexts (like a server that needs to execute a JS script on each incoming request).

The problem is that reusing the same isolate and context several times leads to memory leaks as all allocated values are freed only when the isolate and context are destroyed.

Therefore, it is necessary to free the memory from time to time to avoid leaks. Two functions were added Context.Cleanup() and Isolate.Cleanup().

Cleaning up Isolates and Contexts works only if executed scripts are stateless: they don't use global variables used between executions.

In this example, removing the call to Cleanup() will lead to a memory leak.

```
package main

import (
	"fmt"
	v8 "rogchap.com/v8go"
	"strings"
	"time"
)

func main() {

	// Create an isolate to compile scripts
	source := "console.log(getMsg());"                                              // This script is stateless
	iso1 := v8.NewIsolate()                                                         // Create a new JavaScript VM
	script1, _ := iso1.CompileUnboundScript(source, "test.js", v8.CompileOptions{}) // Compile script to get cached data
	cachedData := script1.CreateCodeCache()

	// Create an isolate to execute scripts
	iso2 := v8.NewIsolate()

	// Create a context with Go callbacks
	global := v8.NewObjectTemplate(iso2)
	_ = global.Set("getMsg", v8.NewFunctionTemplate(iso2, func(info *v8.FunctionCallbackInfo) *v8.Value {
		res, _ := v8.NewValue(iso2, "Hello World!\n")
		return res
	}))
	ctx2 := v8.NewContext(iso2, global)

	// Create a global object to output logs (console.log)
	console := v8.NewObjectTemplate(iso2)
	logfn := v8.NewFunctionTemplate(iso2, func(info *v8.FunctionCallbackInfo) *v8.Value {
		for _, v := range info.Args() {
			fmt.Printf(v.String())
		}
		return nil
	})
	_ = console.Set("log", logfn)
	_ = console.Set("error", logfn)

	for {
		// Reset the objects that have been freed
		consoleObj, _ := console.NewInstance(ctx2)
		_ = ctx2.Global().Set("console", consoleObj)

		// Compile script in new isolate with cached data
		compiledScript, _ := iso2.CompileUnboundScript(source, "test.js", v8.CompileOptions{CachedData: cachedData})
		_, err := compiledScript.Run(ctx2)
		if err != nil {
			panic(err)
		}

		// Cleanup context and isolate to free memory as the script is stateless (no global variables)
		ctx2.Cleanup()
		iso2.Cleanup()
	}
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for compiling a context-dependent UnboundScript which can be run in any context of the isolate it was compiled in.
 - Support for creating a code cache from an UnboundScript which can be used to create an UnboundScript in other isolates
 to run a pre-compiled script in new contexts.
+- Support for freeing isolates and contexts tracked values 
 
 ### Changed
 - Removed error return value from NewIsolate which never fails
--- a/README.md
+++ b/README.md
@@ -106,6 +106,64 @@ script2, _ := iso2.CompileUnboundScript(source, "math.js", v8.CompileOptions{Cac
 val, _ = script2.Run(ctx2)
 ```
 
+
+### Free memory on long-lived isolates and contexts
+
+It is beneficial to create an isolate and a context only once and use them several times instead
+of recreating them each time. Creating an isolate and a context can take 3ms.
+Reusing the same isolate and context several times can lead to memory leaks due to the way the memory
+is managed to wrap v8. Therefore, it is necessary to free the memory from time to time to avoid leaks.
+Long-lived contexts only works if scripts executed are stateless: no global variables used between executions. 
+
+```go
+// Create an isolate to compile scripts
+source := "console.log(getMsg());"                                              // This script is stateless
+iso1 := v8.NewIsolate()                                                         // Create a new JavaScript VM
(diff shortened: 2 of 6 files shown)


dylanahsmith:
A similar proposal was brought up in https://github.com/rogchap/v8go/issues/105.  The problem is that it breaks all memory references V8 objects, which seems like a short term workaround for the problem.  It will leak to crashes if any v8go.Value references from before that Cleanup gets used again.  Using Cleanup properly means holding onto no references, which seems quite limiting.

jblernout:
Of course, but wouldn't that be the responsibility of the developers to make sure they don't use any v8go.Values that have been cleaned up?  It's like using an Isolate that has been disposed.

dylanahsmith:
> Of course, but wouldn't that be the responsibility of the developers to make sure they don't use any v8go.Values that have been cleaned up?

No, I think it should panic instead, since Go code is expected to be memory safe in the absence of data races.  A panic will avoid having undefined behaviour, which could result in security vulnerabilities or at least harder to debug bugs.

Anyways, if we can model a handle scope, then we could support a safer local cleanup, which should support this use case.

PR #234, opened 2021-11-18
Remove Windows support

genevieve:
### Context
The v8go library currently supports binaries for mac, linux, and windows. The last release of this library with an updated version of v8 was in May. We have been making efforts to automate the upgrade process here so that we can ship releases more regularly as patches are released/CVEs are found and fixed. However, we've been unable to merge any of those upgrade PRs opened by the automation @gustavocaso added because it fails to build the window binary on later versions of v8.

The windows binary is built using MinGW and [relies on patches](https://github.com/rogchap/v8go/blob/master/deps/windows_x86_64/README.md) in a [different repository](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-v8) to make the v8 binary compile. These patches have not been updated for later versions and so they fail to apply on upgrade attempts.

Additionally, there seems to be incompatibility with enabling Internationalization/ICU support with the existing windows build process because of what appears to be a reliance on something provided by VSCode.

As such:
1. We cannot ship releases of v8go because we cannot build the windows binary and do not want the versions of mac/linux to be out of step with windows. The last release with a v8 update is from May. We are 6 minors behind.
2. We cannot ship a release of v8go with ICU enabled in the v8 binary because we cannot build the windows binary with it using the current build process.

### Proposal

This PR removes support for windows from v8go as a result of the above issues. If a user of the library can contribute a fix to the above issues now, great. We tried asking in the [#v8go gophers channel](https://gophers.slack.com/archives/C01FW4CR5D2/p1635946889052700) a couple weeks ago but did not get much of a response. Until such time as a user is able to contribute the changes needed to make the windows binary build on recent versions of v8 & with intl support enabled, we can remove it as a blocker allowing us to cut releases of v8go with the latest v8 update for mac and linux, as well as with intl support.


GustavoCaso (approved):

README.md
```diff
@@ -192,21 +192,20 @@ please join the [**#v8go**](https://gophers.slack.com/channels/v8go) channel on
 
 ### Windows
 
-In order to build a project using v8go on Windows, Go requires a gcc compiler to be installed.
+There used to be Windows binary support. For further information see, [this PR](UPDATE).
```
dylanahsmith, line 195: This can be filled in now that the PR is open.

```suggestion
There used to be Windows binary support. For further information see, [PR #234](https://github.com/rogchap/v8go/pull/234).
```

genevieve: Thank you!

README.md
```diff
@@ -247,7 +246,7 @@ The next steps are:
 1) The build is not yet triggered automatically. To trigger it manually, go to the [V8
 Build](https://github.com/rogchap/v8go/actions?query=workflow%3A%22V8+Build%22) Github Action, Select "Run workflow",
 and select your pushed branch eg. `v8_upgrade/<v8-version>`.
-1) Once built, this should open 3 PRs against your branch to add the `libv8.a` for Linux, macOS and Windows; merge
+1) Once built, this should open 2 PRs against your branch to add the `libv8.a` for Linux and macOS; merge
```
dylanahsmith, line 249: Looks like an update was missed on this after ARM64 support was added for macOS

```suggestion
1) Once built, this should open 3 PRs against your branch to add the `libv8.a` for Linux (for x86_64) and macOS for x86_64 and arm64; merge
```

.github/workflows/test.yml
```diff
@@ -45,14 +41,14 @@ jobs:
...
         curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/fossa-cli/master/install.sh | sudo bash
         fossa analyze
       env:
         FOSSA_API_KEY: ${{ secrets.FOSSA_API_KEY }}
     - name: Scan and upload FOSSA data (Windows)
-      if: env.FOSSA_API_KEY != '' && github.ref == 'refs/heads/master' && matrix.platform == 'windows-latest'
+      if: env.FOSSA_API_KEY != '' && github.ref == 'refs/heads/master'
       run: |
         Set-ExecutionPolicy Bypass -Scope Process -Force; iex  ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/fossas/fossa-cli/master/install.ps1'))
         C:\ProgramData\fossa-cli\fossa.exe analyze
```
dylanahsmith, line 54: This windows only step can just be removed now

dylanahsmith (approved):
Approving, but let's wait to hear from @rogchap

epk:
Probably should also delete https://github.com/rogchap/v8go/tree/master/deps/windows_x86_64

rogchap (approved):
Makes sense to me; I've spent too many sleepless nights fighting to get Windows building right, with little success.

dylanahsmith:
> Tests on 1.16.8 windows-latest Expected — Waiting for status to be reported (Required)
> Tests on 1.17.1 windows-latest Expected — Waiting for status to be reported (Required)

@rogchap looks like these will need to be removed from the required checks before merging.

PR #246, opened 2021-12-04
Ignore missing error returns in README

tsak:
This will make examples in README easier to copy and paste. I've seen somebody stumble upon this in the [Gopher Slack](https://gophers.slack.com/archives/C01FW4CR5D2/p1638625454110400) and thought it would probably be helpful if those examples would run out of the box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ import v8 "rogchap.com/v8go"
 ### Running a script
 
 ```go
-ctx := v8.NewContext() // creates a new V8 context with a new Isolate aka VM
+ctx, _ := v8.NewContext() // creates a new V8 context with a new Isolate aka VM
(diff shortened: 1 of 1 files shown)


dylanahsmith:
No, these don't return an error anymore.  As I said on slack

> Look at the documentation for the released version (https://pkg.go.dev/rogchap.com/v8go) instead of the github README which shows the documentation for unreleased code

tsak:
I did test this with a blank project, copy and pasting the code from the README and it was requiring errors either being handled or ignored. As a developer, one probably starts from the README displayed on Github and might try the code examples. I would suggest to have the default branch point to the latest release version so that the default README displayed matches developer expectations and doesn't require a roundtrip to Go docs and/or parsing the library sources.

dylanahsmith:
It is standard for GitHub to default to the latest commit instead of the latest release.

Issue #213, opened 2021-10-26
Add object for-in support (using v8::Object::GetPropertyNames)


dylanahsmith:
This feature request come from a slack question, but seems like something others would want as well

[v8::Object::GetPropertyNames](https://v8.github.io/api/head/classv8_1_1Object.html#a27ea564bd46b89d61bfe7b6c8a399c97) documents itself as providing the same behaviour as the for-in statement:

> The array returned by this method contains the same values as would be enumerated by a for-in statement over this object.

We already have Object.GetIdx that could be used to obtain the values from this Array, although currently reading one past the end of the array would be needed to detect the end fo the array.

snej:
I just ran into this too. I need to get the keys (property names) of an arbitrary object returned from a script, and there's no way to do it. I'm going to have to write a JS wrapper around the script to convert the object into a form I can interpret from Go.

iquirino:
This issue has been addressed in the fork [github.com/iquirino/v8go](https://github.com/iquirino/v8go) (v1.0.0).

The fork is actively maintained, tracks latest stable V8, and includes this fix along with many other improvements. See the [v1 release notes](https://github.com/iquirino/v8go/blob/master/v1.md) for details.

```bash
go get github.com/iquirino/v8go@v1.0.0
```

PR #260, opened 2021-12-23
Add object get property names

GustavoCaso:
Related to https://github.com/rogchap/v8go/issues/213

I'm trying to give a try to my first PR that adds new logic both in Go and in C++.

I'm not 100% sure that want I'm doing is the right thing to do and I would love some guidance if possible. @dylanahsmith @rogchap 

Basically, I'm trying to return an array with all the property names from V8 using the `obj->GetPropertyNames(local_ctx)`. 

Is there anything that I'm missing? 

Thanks 😄


context.go
```diff
@@ -169,6 +169,17 @@ func valueResult(ctx *Context, rtn C.RtnValue) (*Value, error) {
 	return &Value{rtn.value, ctx}, nil
 }
 
+func valueStrings(ctx *Context, rtn C.RtnStrings) []string {
+	length := rtn.length
+	slice := (*[1 << 28]*C.char)(unsafe.Pointer(rtn.strings))[:length:length]
+	var result []string
+	for i := 0; i < len(slice); i++ {
+		s := slice[i]
+		result = append(result, C.GoString(s))
```
GustavoCaso, line 178: Should we be freeing this string?

object.go
```diff
@@ -84,6 +84,13 @@ func (o *Object) SetIdx(idx uint32, val interface{}) error {
 	return nil
 }
 
+// Returns an array containing the names of the enumerable properties of this object,
+// including properties from prototype objects
+func (o *Object) GetPropertyNames() []string {
```
dylanahsmith, line 89: `v8::Object::GetPropertyNames()` returns `MaybeLocal<Array>`, which means that it can fail and that it returns a JS array object instead of a native array for the host.  So I would expect the return value to be `(*Array, error)` in v8go to better match the underlying v8 method, although it looks like we don't yet have an Array type.  The array contents can be retrieved using Object.GetIdx as mentioned in the corresponding issue for this PR (https://github.com/rogchap/v8go/issues/213).

V8 also overloads this method with a more generic version with more control over filtering (this would only return enumerable properties) and key conversion (this would convert indexes to strings).  This more generic version of the method seems like it would be the more appropriate one to expose with this generic `GetPropertyNames` method name.

Given that this is basically exposing a convenience method for returning enumerable properties, perhaps it should be named `GetEnumerablePropertyNames` to better reflect what it returns.  Once viewed in this way as a convenience method, having `([]string, error)` be the return value seems appropriate, since getting a Go slice would be more convenient for typical use and this type would reflect the fact that indexes get converted to strings.

```suggestion
func (o *Object) GetEnumerablePropertyNames() ([]string, error) {
```

v8go.cc
```diff
@@ -1290,6 +1290,27 @@ void ObjectSet(ValuePtr ptr, const char* key, ValuePtr prop_val) {
...
 
+RtnStrings ObjectGetPropertyNames(ValuePtr ptr) {
+  LOCAL_OBJECT(ptr);
+  RtnStrings rtn = {};
+
+  MaybeLocal<Array> maybe_names = obj->GetPropertyNames(local_ctx);
+  Local<Array> names = maybe_names.ToLocalChecked();
+
+  uint32_t length = names->Length();
+  const char** strings = new const char*[length];
```
dylanahsmith, line 1301: This is being deallocated from Go code using C.free which is resulting in an address sanitizer error (`AddressSanitizer: alloc-dealloc-mismatch (operator new [] vs free)`) because it is allocated with the C++ new operator (`new const char*[length]`) then deallocated with C's `free` function.  Instead, this could be allocated using `malloc` to avoid this problem

```suggestion
  const char** strings = malloc(length * sizeof(const char*));
  if (strings == NULL) {
    // TODO: Handle out of memory errors (https://github.com/rogchap/v8go/issues/273)
  }
```

It looks like

v8go.cc
```diff
@@ -1290,6 +1290,27 @@ void ObjectSet(ValuePtr ptr, const char* key, ValuePtr prop_val) {
   obj->Set(local_ctx, key_val, prop_val->ptr.Get(iso)).Check();
 }
 
+RtnStrings ObjectGetPropertyNames(ValuePtr ptr) {
+  LOCAL_OBJECT(ptr);
+  RtnStrings rtn = {};
+
+  MaybeLocal<Array> maybe_names = obj->GetPropertyNames(local_ctx);
+  Local<Array> names = maybe_names.ToLocalChecked();
```
dylanahsmith, line 1298: This could fail if the isolate is terminated from another thread, so we need to use ToLocal instead of ToLocalChecked (e.g. https://github.com/rogchap/v8go/pull/177 fixed a lot of usage of ToLocalChecked)

object.go
GustavoCaso: Fixed here https://github.com/rogchap/v8go/pull/260/commits/19f8b5c1d27a23166f28c79bf792806a5990dc67

v8go.cc
GustavoCaso: Fixed here https://github.com/rogchap/v8go/pull/260/commits/84ff923037accc8f91927c3fc81444845007bdb4

v8go.cc
```diff
@@ -1297,6 +1297,38 @@ void ObjectSet(ValuePtr ptr, const char* key, ValuePtr prop_val) {
...
+    return rtn;
+  }
+
+  uint32_t length = names->Length();
+  const char** strings = (const char**)malloc(length * sizeof(const char*));
+
+  for (uint32_t i = 0; i < length; i++) {
+    Local<Value> name_from_array;
+    if (!names->Get(local_ctx, i).ToLocal(&name_from_array)) {
+      for (i = i - 1; i > 0; i--) {
```
genevieve, line 1316: Will this miss freeing `strings[0]` if the index doesn't get there? Should the middle param be `i >=0`

object_test.go
```diff
@@ -202,6 +203,27 @@ func TestObjectDelete(t *testing.T) {
...
+	obj.Set("bar2", "baz2")
+	obj.Set("foo", "foobar")
+	obj.Set("hello", "world")
+
+	expectedProperties := []string{"bar2", "foo", "hello"}
+	properties, err := obj.GetEnumerablePropertyNames()
+	fatalIf(t, err)
+
+	if !reflect.DeepEqual(properties, expectedProperties) {
+		t.Error("properteis are not the same")
```
genevieve, line 223: properties*

v8go.cc
```diff
@@ -1297,6 +1297,38 @@ void ObjectSet(ValuePtr ptr, const char* key, ValuePtr prop_val) {
...
+  Local<Array> names;
+  if (!obj->GetPropertyNames(local_ctx).ToLocal(&names)) {
+    rtn.error = ExceptionError(try_catch, iso, local_ctx);
+    return rtn;
+  }
+
+  uint32_t length = names->Length();
+  const char** strings = (const char**)malloc(length * sizeof(const char*));
+
+  for (uint32_t i = 0; i < length; i++) {
```
genevieve, line 1313: Question for anyone but when should we be using `uint32_t` vs `int`? [Other for-loops](https://github.com/rogchap/v8go/blob/5e91d3d9dcabd2986f901b6b31590e49fc3c4dd8/v8go.cc#L502) in here use `int`

context.go
```diff
@@ -169,6 +169,22 @@ func valueResult(ctx *Context, rtn C.RtnValue) (*Value, error) {
 	return &Value{rtn.value, ctx}, nil
 }
 
+func valueStrings(ctx *Context, rtn C.RtnStrings) ([]string, error) {
+	if rtn.strings == nil {
+		return []string{}, newJSError(rtn.error)
+	}
+	length := rtn.length
+	slice := (*[1 << 28]*C.char)(unsafe.Pointer(rtn.strings))[:length:length]
+	var result []string
```
genevieve, line 178: Since we know the size of `result`, we should be able to preallocate:

```go
	result := make([]string, len(slice))
	for i := 0; i < len(slice); i++ {
		s := slice[i]
		defer C.free(unsafe.Pointer(s))
		result[i] = C.GoString(s)
	}
	defer C.free(unsafe.Pointer(rtn.strings))
	return result, nil
```

v8go.cc
dylanahsmith: We should try to match the loop bounds to avoid integer overflows.

Using a `uint32_t` makes sense here since [v8::Array::Length()](https://v8.github.io/api/head/classv8_1_1Array.html#ad6b33c7dd543044ca2311c000a1642dd) returns a uint32_t.

The iteration count for the loop you link to uses
[v8::FunctionCallbackInfo::Length()](https://v8.github.io/api/head/classv8_1_1FunctionCallbackInfo.html#af97dd3f1cb01ed039f9479152ad63a84) which returns an `int`, so using an `int` makes the most sense there.

v8go.cc
dylanahsmith: Good catch!

We can't use `i >=0` though, since the integer is unsigned, so that would always be true.  If we wanted to do that, we could change `i` to int64_t.

Would the code maybe be clearer if it matched the outer loop, e.g.

```c++
      length = i;
      for (i = 0; i < length; i++) {
        free(&strings[i]);
```

or by using a separate loop variable?

genevieve: I think this inner loop is supposed to be from `0 to i` whatever `i` is because it might need to free all strings if an error occurs which means that it might not be equal to `length` and we'd be trying to free elements we don't have yet.

genevieve: Maybe it should be:
```
      for (j = 0; j < i; j --) {
        free(&strings[j]);
      }
      free(strings);
      rtn.error = ExceptionError(try_catch, iso, local_ctx);
      return rtn;
    }
```

dylanahsmith: > I think this inner loop is supposed to be from `0 to i`

right, which is my code snippet started with `length = i;`

Your suggestion is what I meant by "or by using a separate loop variable", except that the for loop line needs to declare the `j` variable by prefixing its assignment with its type: `for (uint32_t j = 0; j < i; j++) {`

PR #265, opened 2022-01-05
make sure sure to free the array buffer allocator when calling Isolate Dispose

GustavoCaso:
I was looking at some of the examples files from the v8 API reference and by the looks of it it looks like that they are freeing the array buffer allocator.

https://v8.github.io/api/head/shell_8cc-example.html#a13

I checked and looks like we are not doing that on our end. I added a simple call to `delete iso->GetArrayBufferAllocator();` when we are disposing of the isolate.

### DIFF
--- a/v8go.cc
+++ b/v8go.cc
@@ -185,8 +185,8 @@ void IsolateDispose(IsolatePtr iso) {
     return;
   }
   ContextFree(isolateInternalContext(iso));
-
   iso->Dispose();
+  delete iso->GetArrayBufferAllocator();
 }
 
 void IsolateTerminateExecution(IsolatePtr iso) {


GustavoCaso:
Looks like tests are failing because all the Isolates that we create have the same `default_allocator` for each Isolate. So the moment we free the memory for one Isolate the rest of Isolates would fail to free the memory of the `default_allocator`.

I would imagine that having `default_allocator` being used for all Isolates is the desired behaviour of `v8go`. 

- Should we be concerned knowing that we are not freeing the `default_allocator` memory? @dylanahsmith @rogchap

dylanahsmith:
> * Should we be concerned knowing that we are not freeing the `default_allocator` memory?

No, we don't need to free globals.  In general, freeing memory before the process exits is just wasted work, since the OS already frees the process memory for us after the process exits.

Issue #192, opened 2021-10-01
Avoid null terminated strings in the interface for cgo


dylanahsmith:
Here is a script that demonstrates the problem

```go
package main

import (
	"fmt"
	v8 "rogchap.com/v8go"
)

func main() {
	iso := v8.NewIsolate()
	ctx := v8.NewContext(iso)
	val, err := ctx.RunScript(`"before\x00after"`, "script.js")
	if err != nil { panic(err) }
	fmt.Printf("%#v\n", val.String())
}
```

which prints `"before"` but it should print `"before\x00after"` like `fmt.Printf("%#v\n", "before\x00after")` does.

Similarly, you can try using null characters in JS and you will see that the null character is preserved.

Instead, v8go.h functions should return strings as a character pointer and length (could be wrapped in a struct) so they can be converted to Go strings with `C.GoStringN` and avoid using `C.GoString` for JS strings.  That way these null characters can be preserved.

There also is an opportunity to avoid extra data copying by returning pointers directly to the V8's string data, when we know that data will be copied to a Go string anyways.  Right now, the V8 string data is copied to a null terminated string, which is then copied to a Go string.

backkem:
I'd like to help solve this to make sure the base64 polyfill from [kuoruan/v8go-polyfills](https://github.com/kuoruan/v8go-polyfills) works correctly on '[binary strings](https://developer.mozilla.org/en-US/docs/Web/API/DOMString/Binary)'.

PR #266, opened 2022-01-06
Convert byte array to string

sharmaashish13:
While using a [base64 polyfill](https://github.com/kuoruan/v8go-polyfills/blob/master/base64/base64.go), encountered an issue where passing [binary string](https://developer.mozilla.org/en-US/docs/Web/API/DOMString/Binary) from Go to JS didnt work as expected. 

1. If binary string has a null character in middle of string, CString() function just terminates the value at that null character resulting into a truncated value.
2. v8go.cc's NewValueString function uses NewFromUtf8 which returns invalid response.

Code sample can be found here:
https://github.com/kuoruan/v8go-polyfills/issues/4

Hence, added a new API that uses v8::String::NewFromOneByte to solve this scenario.


value.go
```diff
@@ -53,6 +53,15 @@ func Null(iso *Isolate) *Value {
...
 
+func NewStringFromByteArray(iso *Isolate, val []byte) (*Value, error) {
+	if iso == nil {
+		return nil, errors.New("v8go: failed to create new Value: Isolate cannot be <nil>")
+	}
+	cUint := (*C.uchar)(unsafe.Pointer(&val[0]))
+	rtnVal := C.NewValueStringFromByteArray(iso.ptr, cUint, C.int(len(val)))
+	return valueResult(nil, rtnVal)
+}
+
```
genevieve, line 64: What was your reasoning for introducing a new api instead of adding a switch case to `NewValue` that handles `[]byte` as done below for other types?

value.go
```diff
@@ -53,6 +53,15 @@ func Null(iso *Isolate) *Value {
 	return iso.null
 }
 
+func NewStringFromByteArray(iso *Isolate, val []byte) (*Value, error) {
```
genevieve, line 56: We will probably want the inverse to be supported as well (from the v8 value in JS back to Go)

value.go
sharmaashish13: NewValue converts Go type to V8 type. In future we might want to just implement uint8 array. (there's already a PR related to that implementation). 
But in this scenario, we just want to get a string out of byte array. so created a new api.

value.go
sharmaashish13: As mentioned in another comment, the reason behind implementing byte array to string is to overcome the challenges with null character in binary strings. I'm not sure for which scenario we would want to implement the reverse.

value_test.go
```diff
@@ -473,6 +473,20 @@ func TestValueFunction(t *testing.T) {
...
+	val, err := v8.NewStringFromByteArray(iso, input)
+	if err != nil {
+		t.Errorf("expeced nil but got error %#v", err)
+	}
+	if !val.IsString() {
+		t.Errorf("expeced string but got %s", reflect.TypeOf(val))
+	}
+}
```
GustavoCaso, line 488: Should we be testing that `val` has the correct string value? From the `input` variable is hard to know what is the expected output of calling `NewStringFromByteArray`

value.go
maxshirshin: When v8go converts a `*Value` to a Go `string` (e.g. with `.String()`), it truncates the string at the first null character it encounters.

sharmaashish13: I added a testcase in value_test.go where i am validating the same. If you print val.String(), it doesnt truncate the data even though there is a 0 byte in the input array.
Having said that, do you have any specific test-case? I can look into that scenario.

value.go
maxshirshin: As far as I understand, adding native support for typed arrays in v8go won't help here, because the JS code expects a `string` back, not a typed array. Our issue is that calling `string(bytes)` in Go doesn't create the string we need. That can also be mitigated if we do smth like:

```golang
var s string
for _, b := range bytes {
   s += string(b)
}
```
Then, `s` contains the string we want to pass back to v8.

Because this code implements a special way to handle `[]byte` when converting to an existing v8 type (string), I'd still keep it in `NewValue`.

In any cases, the library user would have to know that doing `string(bytes)` when passing data to V8 leads to erroneous results.

backkem: @maxshirshin based on my understanding the conversion problem, doesn't originate in `string(bytes)`. It seems to be V8's `String::NewStringFromUtf8` that causes the problem: turning valid UTF-8 byte sequences into single characters rather than converting each byte to an equivalent character (as it's supposed to). In this [binary string](https://developer.mozilla.org/en-US/docs/Web/API/DOMString/Binary) case we specifically need 1 byte to result in 1 character. That's why we're using `NewFromOneByte` instead.

I'm not sure how we could implicitly distinguish between the 'binary string' and 'utf-8 string' cases in `NewValue`? As mentioned, passing a `[]byte` would preferably result in a `ArrayBuffer` or `[]Uint8` in the JS realm. If we really want to keep everything inside `NewValue`, we could make a type for it, e.g.: `type BinaryString []byte`. However, the library user would still have to know about this exception case.

value.go
backkem: I guess we could introduce a `(v *Value) BinaryString() []byte` or `(v *Value) BinaryString() BinaryString` depending on [this outcome](https://github.com/rogchap/v8go/pull/266#discussion_r780675085).

backkem: > When v8go converts a `*Value` to a Go `string` (e.g. with `.String()`), it truncates the string at the first null character it encounters.

Note that we're not trying to solve #192 here. We may do that in a separate PR.

value_test.go
sharmaashish13: @GustavoCaso  Updated the testcase to compare output string.

value_test.go
```diff
@@ -473,6 +474,27 @@ func TestValueFunction(t *testing.T) {
...
+	val, err := v8.NewStringFromByteArray(iso, byts)
+	if err != nil {
+		t.Errorf("expected nil but got error %#v", err)
+	}
+	if !val.IsString() {
+		t.Errorf("expected string but got %s", reflect.TypeOf(val))
+	}
+	if val.String() != expected {
+		t.Errorf("expected not same as actual")
+	}
```
dylanahsmith, line 495: Shouldn't this also work for `v8.NewValue(expected)`, in which case this isn't testing the distinguishing behaviour of this new method, which is that the bytes aren't interpreted as UTF-8 in the JS string.  Since there isn't a v8go.String type, a workaround would be to test this in JS (e.g. getting the length and making sure it counts each byte in a UTF-8 encoded byte sequence)

value.go
```diff
@@ -53,6 +53,15 @@ func Null(iso *Isolate) *Value {
...
+	if iso == nil {
+		return nil, errors.New("v8go: failed to create new Value: Isolate cannot be <nil>")
```
dylanahsmith, line 58: This should be treated as a bug, so should instead `panic` to make sure the error is quickly noticed during testing..

value.go
dylanahsmith, line 56: This could use a documentation comment.  A CHANGELOG.md entry would also be appreciated to mention the addition of this new method

dylanahsmith (commented):
Although the test could be improved and documentation could be added, the implementation looks good to me.

value.go
```diff
@@ -53,6 +53,16 @@ func Null(iso *Isolate) *Value {
 	return iso.null
 }
 
+// NewStringFromByteArray returns V8::string from byte array. Converts each byte into equivalent character.
+func NewStringFromByteArray(iso *Isolate, val []byte) (*Value, error) {
```
dylanahsmith, line 57: NewStringFromBytes would be a shorter name.  Referring to the argument as a ByteArray doesn't make it any clearer and could even seem to refer to a JS array before looking at the parameter type.

The parameter name could be named bytes to be consistent with what it is referred to as in the function name.

Also, I've opened https://github.com/rogchap/v8go/pull/277 since this should actually be returning a String type.  We should add that type before adding a constructor for a string, so we don't need a breaking change to change it from Value to String later.

```suggestion
func NewStringFromBytes(iso *Isolate, bytes []byte) (*String, error) {
```

v8go.cc
```diff
@@ -808,6 +808,23 @@ ValuePtr NewValueIntegerFromUnsigned(IsolatePtr iso, uint32_t v) {
   return tracked_value(ctx, val);
 }
 
+RtnValue NewValueStringFromByteArray(IsolatePtr iso, const uint8_t* v, int len){
```
dylanahsmith, line 811: This should be consistent with the Go method it is called from

```suggestion
RtnValue NewStringFromBytes(IsolatePtr iso, const uint8_t* v, int len){
```

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Support for [binary strings](https://developer.mozilla.org/en-US/docs/Web/API/DOMString/Binary). `NewStringFromByteArray` takes bytes array as input and uses `String::NewFromOneByte` to convert each byte to an equivalent character.
```
dylanahsmith, line 10: Binary strings are supported in V8, we are just adding support for constructing it here.  We also shouldn't get into the implementation details of what V8 function it uses here, just briefly describe what is added and enough to make it clear what it is useful for.

```suggestion
- NewStringFromBytes to support the creation of [binary strings](https://developer.mozilla.org/en-US/docs/Web/API/DOMString/Binary) from a Go byte array.
```

value_test.go
```diff
@@ -519,6 +519,59 @@ func TestValueFunction(t *testing.T) {
...
+	t.Parallel()
+	iso := v8.NewIsolate()
+	defer iso.Dispose()
+	global := v8.NewObjectTemplate(iso)
+
+	testFn := v8.NewFunctionTemplate(iso, func(info *v8.FunctionCallbackInfo) *v8.Value {
+		args := info.Args()
+		input := args[0].String()
+		if len(input) == 0 {
+			//String() terminates input at null character. hence hardcoding for 3rd input
```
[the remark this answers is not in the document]
sharmaashish13: .String() not working as expected. for input string `a\x00\xffe`, it should return 4 bytes instead getting 5.

noelcarl:
@dylanahsmith @genevieve is this pull request something we can still move forward with? If not is there another known workaround for supporting the passing of a binary string from Go to JavaScript? Thanks.

cc @sharmaashish13

dylanahsmith:
genevieve and I had been exploring using this library while working at Shopify, which didn't end up adopting this for production use and neither of us work at Shopify anymore.  I think you will need help from someone else with merge access who has been more recently active on this project.

noelcarl:
Ok, sounds good. Thanks for the update and context @dylanahsmith.

PR #277, opened 2022-01-14
Add String type, constructor functions for it and Value.AsString to cast to it

dylanahsmith:
## Problem

https://github.com/rogchap/v8go/pull/266 is attempting to add a constructor function for a V8 string, but we don't yet have a v8go.String type for it to return.  It is also adding a constructor function with a NewString prefix for the edge case of binary strings, but there isn't yet a NewString function for the normal case.  These seem like short-sighted API design decisions that are biased by what is currently available.

## Solution

Add a String type, so that it can be used in APIs to better match the V8 APIs.  We can latter extend this with v8::String methods, but this will still inherit all the Value methods through the embedded Value field.

The NewString function was added to use instead of NewValue, since individual constructors provide better type safety and are more in line with V8's API.

The MustNewString was added as a convenience method for constructing a V8 string from a Go literal string, where we know the function won't fail.

Value.AsString was also added for converting a Value to a String.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Add String type, constructor functions for it and Value.AsString() to cast to it
+
 ### Fixed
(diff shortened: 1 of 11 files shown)


genevieve (approved):
Missing some test coverage otherwise lgtm

PR #281, opened 2022-01-26
Add SnapshotCreator API

GustavoCaso:
This PR allows users of v8go to use the existing API from V8 to create HeapSnapshots.


GustavoCaso:
We could allow passing `SnapshotCreatorOptions` struct from the beginning even if we currently do nothing with it, that way in the future whenever we start adding support for the options on the snapshot creator class we would not break the existing API.

v8go.cc
```diff
@@ -264,6 +276,67 @@ RtnUnboundScript IsolateCompileUnboundScript(IsolatePtr iso,
...
+
+void SetDefaultContext(SnapshotCreatorPtr snapshotCreator, ContextPtr ctx) {
+  Isolate* iso = ctx->iso;
+  Locker locker(iso);
+  Isolate::Scope isolate_scope(iso);
+  HandleScope handle_scope(iso);
+  Local<Context> local_ctx = ctx->ptr.Get(iso);
+  Context::Scope context_scope(local_ctx);
+
+  ContextFree(ctx);
```
GustavoCaso, line 303: The reason we have to free the internal context before adding it into the snapshot creator class is that the snapshot creator class doesn't know how to deserialize the internal values we store in the local context.  Every time we allocate a new `m_value` and pushed to the context `tracked_value(ctx, val)` creates a reference that we have to remove before adding the context.

v8go.cc
```diff
@@ -26,6 +26,7 @@ const int ScriptCompilerEagerCompile = ScriptCompiler::kEagerCompile;
 
 struct m_ctx {
   Isolate* iso;
+  StartupData* startup_data;
```
GustavoCaso, line 29: Adding the `StartupData` to our local context as a convenient location to allow us to free that data once the isolate gets `Dispose()`.

snapshot_creator.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (s *SnapshotCreator) GetIsolate() (*Isolate, error) {
+	if s.ptr == nil {
+		return nil, errors.New("v8go: Cannot get Isolate after creating the blob")
+	}
+
+	return s.iso, nil
+}
+
+func (s *SnapshotCreator) SetDeafultContext(ctx *Context) error {
```
genevieve, line 55: `SetDefaultContext` *

snapshot_creator.go
```diff
@@ -0,0 +1,104 @@
...
+	if s.defaultContextAdded {
+		return errors.New("v8go: Cannot set multiple default context for snapshot creator")
+	}
+
+	C.SetDefaultContext(s.ptr, ctx.ptr)
+	s.defaultContextAdded = true
+	ctx.ptr = nil
```
genevieve, line 62: Instead of the boolean flag on the struct, could we store the v8 context pointer and check if that's nil or not to determine if it's already been set?

context.go
```diff
@@ -76,6 +77,27 @@ func NewContext(opt ...ContextOption) *Context {
 	return ctx
 }
 
+func NewContextFromSnapshot(iso *Isolate, snapshot_index int) (*Context, error) {
```
genevieve, line 80: For all of the new APIs added in this PR, we should have some sort of documentation/comments above each explaining them.

context_test.go
```diff
@@ -39,6 +39,22 @@ func TestContextExec(t *testing.T) {
...
+func TestNewContextFromSnapshotErrorWhenIsolateHasNoStartupData(t *testing.T) {
+	t.Parallel()
+
+	iso := v8.NewIsolate()
+	defer iso.Dispose()
+
+	ctx, err := v8.NewContextFromSnapshot(iso, 1)
+
+	if ctx != nil {
+		t.Errorf("error expected nil context got: %+v", ctx)
```
genevieve, line 51:
```diff
- t.Errorf("error expected nil context got: %+v", ctx)
+ t.Errorf("expected nil context got: %+v", ctx)
```

snapshot_creator.go
```diff
@@ -0,0 +1,104 @@
...
+import "C"
+import (
+	"errors"
+	"unsafe"
+)
+
+type FunctionCodeHandling int
+
+const (
+	FunctionCodeHandlingKlear FunctionCodeHandling = iota
```
genevieve, line 18: I think this should be `FunctionCodeHandlingClear`?

We should add a comment to each of these explaining what they mean (there should be one that already exists where these two parameters are defined in the v8 header files)

snapshot_creator_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	iso := v8.NewIsolate(v8.WithStartupData(data))
+	defer iso.Dispose()
+
+	ctx := v8.NewContext(iso)
+	defer ctx.Close()
+
+	runVal, err := ctx.Global().Get("run")
+	if err != nil {
+		panic(err)
```
genevieve, line 37: I think these if err != nil/panics can be replaced with `fatalIf(t, err)` as above

snapshot_creator_test.go
```diff
@@ -0,0 +1,191 @@
...
+	snapshotCreatorCtx.RunScript(`const add = (a, b) => a + b`, "add.js")
+	snapshotCreatorCtx.RunScript(`function run() { return add(3, 4); }`, "main.js")
+	err = snapshotCreator.SetDeafultContext(snapshotCreatorCtx)
+	fatalIf(t, err)
+
+	err = snapshotCreator.SetDeafultContext(snapshotCreatorCtx)
+	defer snapshotCreatorCtx.Close()
+
+	if err == nil {
+		t.Error("Adding an extra default cointext show have fail")
```
genevieve, line 147:
```suggestion
		t.Error("setting another default context should have failed, got <nil>")
```

v8go.cc
```diff
@@ -264,6 +276,67 @@ RtnUnboundScript IsolateCompileUnboundScript(IsolatePtr iso,
...
+
+  ContextFree(ctx);
+
+  return snapshotCreator->AddContext(local_ctx);
+}
+
+RtnSnapshotBlob* CreateBlob(SnapshotCreatorPtr snapshotCreator,
+                            int function_code_handling) {
+  //  kKeep - keeps any compiled functions
+  //  kClear - does not keep any compiled functions
```
genevieve, line 324: Could pull these comments out to the go file

v8go.cc
```diff
@@ -152,9 +153,19 @@ void Init() {
   return;
 }
 
-IsolatePtr NewIsolate() {
+IsolatePtr NewIsolate(IsolateOptions options) {
   Isolate::CreateParams params;
   params.array_buffer_allocator = default_allocator;
+
+  StartupData* startup_data;
+  if (options.snapshot_blob_data != nullptr) {
```
genevieve, line 161: I think this applies here: https://stackoverflow.com/a/17772117
```suggestion
  if (options.snapshot_blob_data) {
```

snapshot_creator.go
GustavoCaso: I thought about that at first, but the problem with using that approach is that after setting the default context we have to reset its pointer to `nil` because in `v8go.cc` we already free it. So having to set the pointer to `nil` after will make it so that the check fo `s.defaultCtx == nil` will always be true.

GustavoCaso:
@rogchap If you have sometime I would appreciate your feedback on the new API for snapshot creation 😄

rogchap:
On a very quick glance, the API remains very close to the CPP equivalent, and remains mostly backward compatible which is great.

I'll need to spend a little more time to look at this in detail. 

Appreciate your contributions 💪

GustavoCaso:
Hi @rogchap 👋 

By any chance have you got some time to look into it?

context_test.go
```diff
@@ -39,6 +39,61 @@ func TestContextExec(t *testing.T) {
...
+	defer iso.Dispose()
+
+	ctx, err := v8.NewContextFromSnapshot(iso, index+1)
+
+	if ctx != nil {
+		t.Errorf("expected nil context got: %+v", ctx)
+	}
+	if err == nil {
+		t.Error("error expected but was <nil>")
+	}
```
genevieve, line 94: Does it make sense to test that the functions from the snapshots are present in this new isolate? i.e. `run` and `multiply` are available? Note that the two contexts in the snapshot both set a `function run()` but the first is performing an add and the second i performing a multiplication so I'm not sure which wins.

isolate.go
```diff
@@ -44,20 +45,48 @@ type HeapStatistics struct {
...
 // different isolates simultaneously.
 // When an isolate is no longer used its resources should be freed
 // by calling iso.Dispose().
+// If StartupData is passed as part of createOptions it will be use as part
+// of the createParams. The StartupData will be use when creating new context
+// from the Isolate.
 // An *Isolate can be used as a v8go.ContextOption to create a new
 // Context, rather than creating a new default Isolate.
-func NewIsolate() *Isolate {
+func NewIsolate(opts ...createOptions) *Isolate {
```
genevieve, line 70: We should update the CHANGELOG to reflect the changes in this PR

snapshot_creator.go
```diff
@@ -0,0 +1,116 @@
...
+// #include "v8go.h"
+import "C"
+import (
+	"errors"
+	"unsafe"
+)
+
+type FunctionCodeHandling int
+
+//  Clear - does not keeps any compiled data prior to serialization/deserialization/verify pass
```
genevieve, line 17: keep*

snapshot_creator.go
```diff
@@ -0,0 +1,116 @@
...
+
+	return &SnapshotCreator{
+		ptr:                 rtn.creator,
+		iso:                 &Isolate{ptr: rtn.iso},
+		defaultContextAdded: false,
+	}
+}
+
+// GetIsolate returns the Isolate associated with the SnapshotCreator.
+// This Isolate must be use to create the contexts that later will be use to create the snapshot blob.
```
genevieve, line 53: used* in 2x

snapshot_creator.go
```diff
@@ -0,0 +1,116 @@
...
+func (s *SnapshotCreator) GetIsolate() (*Isolate, error) {
+	if s.ptr == nil {
+		return nil, errors.New("v8go: Cannot get Isolate after creating the blob")
```
genevieve, line 56: Might be worth panicking here trying to call a function on a snapshot creator thats likely been disposed of (either through `Dispose` or from calling `Create` to make the blob)

snapshot_creator.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	if !s.defaultContextAdded {
+		return nil, errors.New("v8go: Cannot create a snapshot without a default context")
+	}
+
+	rtn := C.CreateBlob(s.ptr, C.int(functionCode))
+
+	s.ptr = nil
+	s.iso.ptr = nil
```
genevieve, line 101: Does the isolate need to be explicitly disposed with `s.iso.Dispose()`?

snapshot_creator.go
```diff
@@ -0,0 +1,116 @@
...
+	C.SnapshotBlobDelete(rtn)
+
+	return &StartupData{data: data, raw_size: raw_size}, nil
+}
+
+// Dispose deletes the reference to the SnapshotCreator.
+func (s *SnapshotCreator) Dispose() {
+	if s.ptr != nil {
+		C.DeleteSnapshotCreator(s.ptr)
+	}
```
genevieve, line 115: Should this set `s.ptr = nil` once its been deleted?

context_test.go
genevieve: Nvm, I see that it's covered here: https://github.com/rogchap/v8go/pull/281/files#diff-d3c12fa8121da12a780c6095ddbc336ce58bcc8714068213feefa96da45f181f

We could probably simplify the scripts set on these two contexts since theyre not being tested.

snapshot_creator_test.go
```diff
@@ -0,0 +1,180 @@
...
+	fatalIf(t, err)
+	snapshotCreatorCtx := v8.NewContext(snapshotCreatorIso)
+
+	snapshotCreatorCtx.RunScript(`const add = (a, b) => a + b`, "add.js")
+	snapshotCreatorCtx.RunScript(`function run() { return add(3, 4); }`, "main.js")
+	err = snapshotCreator.SetDefaultContext(snapshotCreatorCtx)
+	fatalIf(t, err)
+
+	err = snapshotCreator.SetDefaultContext(snapshotCreatorCtx)
+	defer snapshotCreatorCtx.Close()
```
genevieve, line 133: Should this defer move up to line 125 where the context is created? In case theres an error in a prior step, we'd still want this cleaned up.

snapshot_creator_test.go
```diff
@@ -0,0 +1,180 @@
...
+		t.Error("adding context should have fail")
+	}
+
+	_, err = snapshotCreator.Create(v8.FunctionCodeHandlingClear)
+	if err == nil {
+		t.Error("creating snapshot should have fail")
+	}
+}
+
+func TestCreateSnapshotErrorIfNodefaultContextIsAdded(t *testing.T) {
```
genevieve, line 171: NoDefault*

v8go.cc
```diff
@@ -593,9 +664,47 @@ ContextPtr NewContext(IsolatePtr iso,
...
+  // the complexities of C -> Go function pointers, we store a reference to the
+  // context as a simple integer identifier; this can then be used on the Go
+  // side to lookup the context in the context registry. We use slot 1 as slot 0
+  // has special meaning for the Chrome debugger.
+
+  Local<Context> local_ctx;
+  MaybeLocal<Context> maybe_local_ctx =
+      Context::FromSnapshot(iso, snapshot_blob_index);
+
+  if (!maybe_local_ctx.ToLocal(&local_ctx)) {
```
genevieve, line 690:
```diff
-  MaybeLocal<Context> maybe_local_ctx = Context::FromSnapshot(iso, snapshot_blob_index);
+  if (!Context::FromSnapshot(iso, snapshot_blob_index).ToLocal(&local_ctx)) {
```

snapshot_creator.go
```diff
@@ -0,0 +1,116 @@
...
+		return 0, errors.New("v8go: Cannot add context to snapshot creator after creating the blob")
+	}
+
+	index := C.AddContext(s.ptr, ctx.ptr)
+	ctx.ptr = nil
+
+	return int(index), nil
+}
+
+// Create creates a snapshot data blob.
```
genevieve, line 88: Might want to mention that the snapshot creator becomes unusable after since it is disposed of.

snapshot_creator.go
genevieve: Ah, since `CreateBlob` deletes the snapshot creator, it already deletes its own isolate so there's no need to dispose of both explicitly.
> deps/v8/include/v8-snapshot.h
>    * Destroy the snapshot creator, and exit and dispose of the Isolate associated with it.

snapshot_creator.go
GustavoCaso: Nice catch 😄

snapshot_creator_test.go
GustavoCaso: I'm not sure if the location of the `defer` matters in this case. Since the test would exit and execute all the defer statements. I do agree that having the defer a few lines below where the variable is created seems unnecessary. I'm moving it up

snapshot_creator.go
GustavoCaso: It seems that the pattern among other functions https://github.com/rogchap/v8go/pull/118#discussion_r619480410

Should we panic in all of the instances. It seems that it would make working with the Snapshot creator API much easier. 

```golang
snapshotCreator := v8.NewSnapshotCreator()
snapshotCreatorIso := snapshotCreator.GetIsolate()

snapshotCreatorCtx := v8.NewContext(snapshotCreatorIso)
defer snapshotCreatorCtx.Close()

snapshotCreatorCtx.RunScript(`const add = (a, b) => a + b`, "add.js")
snapshotCreatorCtx.RunScript(`function run() { return add(3, 4); }`, "main.js")
snapshotCreator.SetDefaultContext(snapshotCreatorCtx)
data := snapshotCreator.Create(v8.FunctionCodeHandlingClear)

iso := v8.NewIsolate(v8.WithStartupData(data))
defer iso.Dispose()

ctx, err := v8.NewContextFromSnapshot(iso, index)
defer ctx.Close()
```

The only one that I'm not sure 100% we should panic is the one that errors when calling `Create` before adding a default context since this is something a developer can fix. 

What do you think @genevieve?

genevieve: Lets stick the error

genevieve (approved):

snapshot_creator.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil, errors.New("v8go: Cannot create a snapshot without a default context")
+	}
+
+	rtn := C.CreateBlob(s.ptr, C.int(functionCode))
+
+	s.ptr = nil
+	s.iso.ptr = nil
+
+	raw_size := rtn.raw_size
+	data := C.GoBytes(unsafe.Pointer(rtn.data), raw_size)
```
dylanahsmith, line 105: The v8::SnapshotCreator::CreateBlob documentation says it returns "{ nullptr, 0 } on failure".  If this function is going to return an error, then I would definitely expect an error to be returned when V8 would return an error.  Unfortunately, V8 is quite vague on what possible reasons there are for an error and the API doesn't expose an error message.

snapshot_creator.go
```diff
@@ -0,0 +1,118 @@
...
+
+	index := C.AddContext(s.ptr, ctx.ptr)
+	ctx.ptr = nil
+
+	return int(index), nil
+}
+
+// Create creates a snapshot data blob.
+// The snapshot creator instance is unsable after creating the snapshot data blob.
+func (s *SnapshotCreator) Create(functionCode FunctionCodeHandling) (*StartupData, error) {
```
dylanahsmith, line 90: StartupData is basically just a wrapper around a slice.  It may as well be returned without the extra pointer indirection, since the error would indicate whether or not it is an error.  Since V8 doesn't actually expose an error message, `ok` could be used for the error and the `s.ptr == nil` and `!s.defaultContextAdded` cases could result in a panic.

Danm87:
Hi,
this is something i'm interested in as well. 
would appreciate to know what's the current blocker :)

PR #285, opened 2022-02-10
Expose HitCount and HitLineCount to CPUProfileNode

mnutt:
This exposes hitCount and hitLineCount from v8's cpuprofile node. It's useful in order to be able to see relative percentages of time spent in a v8 function and its descendants, although computing total time is left up to the presentation implementation.

Sample implementation:

```go
func printTree(nest string, node *v8.CPUProfileNode) {
	file, name, line, col := node.GetScriptResourceName(), node.GetFunctionName(), node.GetLineNumber(), node.GetColumnNumber()
	selfHits, totalHits := node.GetHitCount(), totalHitCountForNode(node)

	fmt.Printf("[%d/%d hits\t]  %s%s() %s:%d:%d\n", selfHits, totalHits, nest, name, file, line, col)

	count := node.GetChildrenCount()
	if count == 0 {
		return
	}

	nest = fmt.Sprintf("%s  ", nest)

	for i := 0; i < count; i++ {
		printTree(nest, node.GetChild(i))
	}
}

func totalHitCountForNode(node *v8.CPUProfileNode) uint {
	totalHitCount := node.GetHitCount()

	count := node.GetChildrenCount()
	for i := 0; i < count; i++ {
		totalHitCount += totalHitCountForNode(node.GetChild(i))
	}

	return totalHitCount
}
```

### DIFF
--- a/cpuprofilenode.go
+++ b/cpuprofilenode.go
@@ -17,6 +17,12 @@ type CPUProfileNode struct {
 	// The number of the column where the function originates.
 	columnNumber int
 
+	// The number of samples recorded in the function
+	hitCount uint
+
+	// The number of lines in which samples were recorded in the function
+	hitLineCount uint
+
 	// The children node of this node.
 	children []*CPUProfileNode
 
@@ -42,6 +48,16 @@ func (c *CPUProfileNode) GetLineNumber() int {
 // Returns number of the column where the function originates.
 func (c *CPUProfileNode) GetColumnNumber() int {
 	return c.columnNumber
+}
+
+// Returns total samples recorded inside the function
+func (c *CPUProfileNode) GetHitCount() uint {
+	return c.hitCount
+}
+
+// Returns total number of lines inside the function recording samples
+func (c *CPUProfileNode) GetHitLineCount() uint {
+	return c.hitLineCount
 }
 
 // Retrieves the ancestor node, or nil if the root.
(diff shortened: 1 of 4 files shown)


cpuprofilenode.go
```diff
@@ -17,6 +17,12 @@ type CPUProfileNode struct {
 	// The number of the column where the function originates.
 	columnNumber int
 
+	// The number of samples recorded in the function
+	hitCount uint
```
genevieve, line 21: Would you be open to using the comment from the v8 header file for these?

i.e.
```
// The count of samples where the function was currently executing.
hitCount uint
```

```
// The number of the function's source lines that collect the samples.
hitLineCount uint
```

cpuprofilenode.go
```diff
@@ -44,6 +50,16 @@ func (c *CPUProfileNode) GetColumnNumber() int {
...
 
+// Returns total samples recorded inside the function
+func (c *CPUProfileNode) GetHitCount() uint {
+	return c.hitCount
+}
+
+// Returns total number of lines inside the function recording samples
+func (c *CPUProfileNode) GetHitLineCount() uint {
+	return c.hitLineCount
+}
```
genevieve, line 61: Same for the getters ^

genevieve:
The changes look good.
1. Could you update the changelog to reflect these two new functions on nodes?
2. Are there test assertions in the node suite that could cover these two new functions while also being reliable?

PR #289, opened 2022-02-22
Add more samples in README

Laisky:
add two useful samples in readme

1. call go func in js and return some values
2. resolve js promise in go


README.md
```diff
@@ -40,21 +40,36 @@ if _, err := ctx2.RunScript("multiply(3, 4)", "main.js"); err != nil {
...
 global := v8.NewObjectTemplate(iso) // a template that represents a JS Object
-global.Set("print", printfn) // sets the "print" property of the Object to our function
+global.Set("add", addFn) // sets the "add" property of the Object to our function
 ctx := v8.NewContext(iso, global) // new Context with the global Object set to our object template
-ctx.RunScript("print('foo')", "print.js") // will execute the Go callback with a single argunent 'foo'
+ret, _ := ctx.RunScript("add(1, 2)", "add.js") // will execute the Go callback with a single argunent 'foo'
+fmt.Println(ret) // Output: 3
 ```
 
+### Resolve JavaScript Promise in Go
```
dylanahsmith, line 61: This example isn't resolving the promise in Go, it is just getting the result.

README.md
```diff
@@ -40,21 +40,36 @@ if _, err := ctx2.RunScript("multiply(3, 4)", "main.js"); err != nil {
...
-### JavaScript function with Go callback
+### Call Go function from JavaScript
 
 ```go
 iso := v8.NewIsolate() // create a new VM
 // a template that represents a JS function
-printfn := v8.NewFunctionTemplate(iso, func(info *v8.FunctionCallbackInfo) *v8.Value {
-    fmt.Printf("%v", info.Args()) // when the JS function is called this Go callback will execute
-    return nil // you can return a value back to the JS caller if required
+addFn := v8.NewFunctionTemplate(iso, func(info *v8.FunctionCallbackInfo) *v8.Value {
```
dylanahsmith, line 48: Why was this example changed?

Laisky: > Why was this example changed?

add  sample `return real value to js`

genevieve:
I wonder if it's more practical in the long term to keep the examples in the readme minimal. The tests cover almost all of the common usecases including [returning a value in the function template](https://github.com/rogchap/v8go/blob/5e91d3d9dcabd2986f901b6b31590e49fc3c4dd8/function_template_test.go#L61-L65), [handling the arguments to a function](https://github.com/rogchap/v8go/blob/5e91d3d9dcabd2986f901b6b31590e49fc3c4dd8/function_template_fetch_test.go#L28-L29), and [getting the returned val as a promise](https://github.com/rogchap/v8go/blob/7df088a0dbd185b3f7c3ef0989876859cf6baed5/function_template_fetch_test.go#L45-L46). The benefit of leaning on tests is that they have to compile and work for the code on a given commit (in order to pass CI) so the tests will reflect the expected usage of the code. Markdown can more easily get out of date.

PR #325, opened 2022-06-21
Reduce frequency for automatic opening of V8 upgrade PRs

dylanahsmith:
## Problem

I noticed that there were 11 open "Upgrade V8 binaries" PRs which had been automatically opened and had been sitting for over a month.

## Solution

Reduce the frequency for automatically opening these PRs to once a month rather than once a day.  We can always increase or decrease the frequency going forward, but this seems like it would be closer to the current frequency that V8 is bumped in practice.

### DIFF
--- a/.github/workflows/v8upgrade.yml
+++ b/.github/workflows/v8upgrade.yml
@@ -3,7 +3,7 @@ name: V8 Upgrade
 on:
   workflow_dispatch:
   schedule:
-    - cron: '0 0 * * *' # Run every day
+    - cron: '0 0 1 * *' # Run every day
 
 jobs:
     upgrade:


dylanahsmith:
@jacques-n any thoughts on how frequently you think we should update the V8 binaries?

jacques-n:
I think the frequency is really a red herring. The PR is only opened when there is an actual change so there is no need to match the frequency of check with the frequency of releases, that happens automatically.  If the project maintainers hadn't been idle, I'd suspect that we would have had this (as I don't think the frequency is terribly burdensome). Beyond that, if the behavior was similar to dependabot, I don't think the frequency would be an issue. Specifically, if we changed the following
1. a single PR that is complete and requires no user intervention beyond merging (as opposed to the current pattern of multiple steps of user engagement + intermediate waiting to get something done: trigger build, merge 4 prs for builds once done, then merge base pr)
2. old update PRs are replaced with new updates PRs (or old are closed when new are created)

I think that if we made the changes above, this would be a non-issue. I'm hoping I'll at least be able to pick up item 1 above as I think that is pretty straightforward.

jacques-n:
Just to clarify, my initial sounds like it is negative on changing this. That wasn't my intention. Until we resolve the other issues I mentioned, this seems like a fine change to me.

jacques-n (approved):

dylanahsmith:
Note that it looks like I'm the one being notified of failures for the scheduled V8 Upgrade workflow.

Ideally, someone else could receive these notifications who won't just ignore them.  https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs suggests a couple of ways that can be done

> If a different user updates the cron syntax in the workflow file, subsequent notifications will be sent to that user instead. If a scheduled workflow is disabled and then re-enabled, notifications will be sent to the user who re-enabled the workflow rather than the user who last modified the cron syntax.

PR #356, opened 2022-12-19
Support []byte in Value

fizx:
This is a reduction of the longstanding #143 to do the minimum I actually need: letting me put []byte's into and out of values.

### DIFF
--- a/v8go.cc
+++ b/v8go.cc
@@ -830,6 +830,46 @@ RtnValue NewValueString(IsolatePtr iso, const char* v, int v_length) {
   return rtn;
 }
 
+ValuePtr NewValueUint8Array(IsolatePtr iso, const uint8_t *v, int len) {
+  ISOLATE_SCOPE_INTERNAL_CONTEXT(iso);
+
+  Local<Context> local_ctx = ctx->ptr.Get(iso);
+  Context::Scope context_scope(local_ctx);
+
+  std::unique_ptr<BackingStore> bs = ArrayBuffer::NewBackingStore(
+    static_cast<void*>(const_cast<uint8_t*>(v)), len,
+    [](void* data, size_t length, void *deleter_data) {
+      free(data);
+      }, nullptr);
+
+  Local<ArrayBuffer> arbuf = ArrayBuffer::New(iso, std::move(bs));
+
+  m_value* val = new m_value;
+  val->iso = iso;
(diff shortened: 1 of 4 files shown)


RGood (approved):
Excellent changes 👍 🚀

v8go.cc
```diff
@@ -830,6 +830,46 @@ RtnValue NewValueString(IsolatePtr iso, const char* v, int v_length) {
...
+ValuePtr NewValueUint8Array(IsolatePtr iso, const uint8_t *v, int len) {
+  ISOLATE_SCOPE_INTERNAL_CONTEXT(iso);
+
+  Local<Context> local_ctx = ctx->ptr.Get(iso);
+  Context::Scope context_scope(local_ctx);
+
+  std::unique_ptr<BackingStore> bs = ArrayBuffer::NewBackingStore(
+    static_cast<void*>(const_cast<uint8_t*>(v)), len,
+    [](void* data, size_t length, void *deleter_data) {
+      free(data);
```
niedzielski, line 842: As far as I can tell, this seems to be the correct deallocator to use but I couldn't find a convincing example.

v8go.cc
```diff
@@ -830,6 +830,46 @@ RtnValue NewValueString(IsolatePtr iso, const char* v, int v_length) {
...
+  int length = array.ToLocalChecked()->ByteLength();
+  uint8_t* bytes = new uint8_t[length];
+  memcpy(bytes, array.ToLocalChecked()->Buffer()->GetBackingStore()->Data(), length);
+  return bytes;
+}
+
+// Returns length of the array (number of elements, not number of bytes)
+uint64_t ValueToArrayLength(ValuePtr ptr) {
+  LOCAL_VALUE(ptr);
+  MaybeLocal<TypedArray> array = value.As<TypedArray>();
```
niedzielski, line 869: Do we need to check if this is empty?

```
uint64_t ValueToArrayLength(ValuePtr ptr) {
  LOCAL_VALUE(ptr);
  MaybeLocal<TypedArray> array = value.As<TypedArray>();
  if (array.IsEmpty()) {
    return 0;
  }
  return array.ToLocalChecked()->Length();
}
```

Same question above in ValueToUint8Array().

niedzielski (changes requested):

PR #361, opened 2023-01-13
v8.Value becomes manually releaseable

fizx:
This allows long-lived contexts that call go functions without leaking memory.  Typically, you'd want to use the following pattern for go functions that don't allow values to escape.

	printfn := v8.NewFunctionTemplate(iso, func(info *v8.FunctionCallbackInfo) *v8.Value {
		defer info.Release()
		fmt.Printf("%+v\n", info.Args())
		return nil
	})

### DIFF
--- a/function_template_test.go
+++ b/function_template_test.go
@@ -48,6 +48,44 @@ func TestFunctionTemplate_panic_on_nil_callback(t *testing.T) {
 	defer iso.Dispose()
 	v8.NewFunctionTemplate(iso, nil)
 }
+func TestFunctionTemplate_generates_values(t *testing.T) {
+	t.Parallel()
+
+	iso := v8.NewIsolate()
+	defer iso.Dispose()
+	global := v8.NewObjectTemplate(iso)
+	printfn := v8.NewFunctionTemplate(iso, func(info *v8.FunctionCallbackInfo) *v8.Value {
+		fmt.Printf("%+v\n", info.Args())
+		return nil
+	})
+	global.Set("print", printfn, v8.ReadOnly)
+	ctx := v8.NewContext(iso, global)
+	defer ctx.Close()
+	ctx.RunScript("print('foo', 'bar', 0, 1)", "")
+	if ctx.RetainedValueCount() != 5 {
+		t.Errorf("expected 5 retained values, got: %d", ctx.RetainedValueCount())
+	}
+}
+
+func TestFunctionTemplate_releases_values(t *testing.T) {
+	t.Parallel()
+
+	iso := v8.NewIsolate()
+	defer iso.Dispose()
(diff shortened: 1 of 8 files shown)


RGood (approved):
This is great if it can be used to mitigate the memleak issues associated with https://github.com/rogchap/v8go/issues/105

The code looks good, but it would be nice to get a memory profile demonstrating it

v8go.cc
```diff
@@ -26,12 +27,14 @@ const int ScriptCompilerEagerCompile = ScriptCompiler::kEagerCompile;
...
   Isolate* iso;
-  std::vector<m_value*> vals;
+  std::unordered_map<long, m_value*> vals;
   std::vector<m_unboundScript*> unboundScripts;
   Persistent<Context> ptr;
+  long nextValId;
 };
 
 struct m_value {
+  long id;
```
niedzielski, line 37: I think we need a constructor here to enforce that this member is initialized to zero (eg, `m_value() : id(0) { }`). Otherwise, we have to depend on the caller to invoke like `new m_value()` instead of `new m_value` and everybody here is doing the latter. The reason I think this is that we explicitly check whether id is zero further down.

v8go.cc
```diff
@@ -600,16 +606,22 @@ ContextPtr NewContext(IsolatePtr iso,
...
 
-  for (m_value* val : ctx->vals) {
-    val->ptr.Reset();
-    delete val;
+  
+  for(auto& [key, value] : ctx->vals) {
+    value->ptr.Reset();
+    delete value;
   }
+  ctx->vals.clear();
```
niedzielski, line 624: :no_bell: I don't know about perf but I think it'd be clearer to call ValueRelease() (but be careful about modifying while looping).

niedzielski (changes requested):

function_template_test.go
```diff
@@ -48,6 +48,45 @@ func TestFunctionTemplate_panic_on_nil_callback(t *testing.T) {
...
 	v8.NewFunctionTemplate(iso, nil)
 }
+func TestFunctionTemplate_generates_values(t *testing.T) {
+	t.Parallel()
+
+	iso := v8.NewIsolate()
+	defer iso.Dispose()
+	global := v8.NewObjectTemplate(iso)
+	printfn := v8.NewFunctionTemplate(iso, func(info *v8.FunctionCallbackInfo) *v8.Value {
+		fmt.Printf("%+v\n", info.Args())
```
niedzielski, line 58: Just to keep the test output clean, please consider removing any unneeded print statements here. I know the test is trying to show a print implementation but we should probably check the value here, not print it out. Same thing below.

niedzielski (approved):
:1st_place_medal:

PR #363, opened 2023-01-14
Upgrade V8 binaries for 10.9.194.9 version **working**

jacques-n:
Use existing workflows to build latest V8 artifacts as well as code changes to ensure v8go continues to work.

Includes:

- Update github workflow to use go 1.8 & 1.9
- Update github workflow to use macos-latest
- Update github build workflow to use ubuntu 22.04
- Add gitignore for jetbrains and .gclient_previous files
- Switch cgo build to C++17 and enable sandbox at build time
- Update test with update to date error message
- Remove no longer supported build flag.
- Move initialization to v8go.go and include flag set to avoid flag freezing
- Reorder initialization so allocator is initialized after v8 (required by latest v8)
- Commits are kept separate with merges to certify provenance of all library objects built using existing rogchap/v8go mechanisms.

This is reopen of the previous PR using a different branch name so github action automation doesn't accidentally close or overwrite.


jacques-n:
@fizx, can you review this? It's basically trivial changes other than the v8 upgrade.

deps/build.py
```diff
@@ -67,7 +67,6 @@
 v8_enable_i18n_support=true
 icu_use_data_file=false
 v8_enable_test_features=false
-v8_untrusted_code_mitigations=false
```
RGood, line 70: What does removing this do?

jacques-n: Removes an error (warning) about the use of a non-existent build flag. I see old references to it in cloudflare blogs but as far as I can tell, it is no longer valid in v8. It seems like they replaced all that code with a cleaner sandboxing model (which we have enabled-- you'll see that flag added elsewhere). Unfortunately, it seems like the v8 developers aren't super transparent about these kinds of things.

RGood (approved):
Last ask I have would be to update the `CHANGELOG.md`

dylanahsmith (changes requested):
https://github.com/rogchap/v8go#v8-dependency says

> Due to security concerns of binary blobs hiding malicious code, the V8 binary is built via CI ONLY.

Could you instead open a PR against https://github.com/rogchap/v8go/pull/357 with the changes needed to fix the build, but without committing the actual V8 build.  I can then merge this PR into that branch and trigger the V8 Build workflow (as documented in https://github.com/rogchap/v8go#upgrading-the-v8-binaries) to merge in the V8 binaries.

jacques-n:
> https://github.com/rogchap/v8go#v8-dependency says
> 
> > Due to security concerns of binary blobs hiding malicious code, the V8 binary is built via CI ONLY.
> 
> Could you instead open a PR against the #357 branch with the changes needed to fix the build, but without committing the actual V8 build. I can then merge this PR into that branch and trigger the V8 Build workflow (as documented in https://github.com/rogchap/v8go#upgrading-the-v8-binaries) to merge in the V8 binaries.

Can do. 

FYI: this patch already followed that pattern to address this concern: the v8 build commits within this patch are were generated by Github action CI and are maintained as separate commits to guarantee hash provenance. (You'll need to look at the individual commits of this PR to actually see this-- each is a GH generated PR with merge commit to guarantee hash history)

dylanahsmith:
Sorry, my mistake about thinking the build wasn't CI generated.  Github Actions seems to now attribute the automated commits to a person (e.g. https://github.com/rogchap/v8go/pull/357/commits/208b63ea0bd4c82f904e0029bafa32a4c0197f0e is attributed to me) which made it look like a local build.

.github/workflows/test.yml
```diff
@@ -12,11 +12,11 @@ jobs:
     name: Tests on ${{ matrix.go-version }} ${{ matrix.platform }}
     strategy:
       matrix:
-        go-version: [1.16.8, 1.17.1]
+        go-version: [1.18.10, 1.19.5]
```
dylanahsmith, line 15: Looks like the PR is currently blocked on required commit statuses that are used along with the repo's branch protection rules

> Tests on 1.16.8 macos-11 Expected — Waiting for status to be reported
> Tests on 1.16.8 ubuntu-latest Expected — Waiting for status to be reported
> Tests on 1.17.1 macos-11 Expected — Waiting for status to be reported
> Tests on 1.17.1 ubuntu-latest Expected — Waiting for status to be reported

Looks like we will need @rogchap to update these in order to allow the to be merged, since contributors like myself don't have access to changes settings like these.

In order to not have to update these settings each time the go version is updated, we could give more stable names for the go version.  E.g.

```
    name: Tests with ${{ matrix.go.name }} Go on ${{ matrix.platform }}
    strategy:
      matrix:
        go:
          - name: 'minimum supported'
            version: '1.16.8'
          - name: 'latest'
            version: '1.17.1'
```

will give names like "Tests with minimum supported Go on ubuntu-latest", so we could more easily bump these versions when a new version is released or when a version becomes end-of-life.

rogchap: I've updated the required checks. All green now.

dylanahsmith (approved):

jacques-n:
> Sorry, my mistake about thinking the build wasn't CI generated. Github Actions seems to now attribute the automated commits to a person (e.g. [208b63e](https://github.com/rogchap/v8go/commit/208b63ea0bd4c82f904e0029bafa32a4c0197f0e) is attributed to me) which made it look like a local build.

I've kept as is.

If you'd like, I can back out the v8 builds but then someone else just has to do them again :D

I've updated the changelog. Let me know if you are happy with this.

dylanahsmith:
> This branch is out-of-date with the base branch

Looks like the branch protection requires branches to be up to date before merging.  I also don't have permission to push a merge of master into this branch to fix it myself for merging.  @jacques-n could you get this branch up-to-date so it can be merged?

jacques-n:
Updated

PR #391, opened 2023-08-19
Use profiles to track object allocation and release.

ajaysusarla:
Use profiles to track allocation and release of objects within go, based on build tags.


ajaysusarla:
I've created an issue to track this work: https://github.com/rogchap/v8go/issues/392

